PR #162, opened 2013-10-30
Add .select method to deffered and pass it as criteria

academo:
I made this method to allow call .select in a find object to allow select which field would be retrieved. Each adapter must implement the select feature by its own which is passed in criteria. If not implement in adapter the method won't do anything.


particlebanana:
merging with #509

Issue #134, opened 2013-10-01
soft delete (on attributes)


mikermcneil:
no idea how this should be set up-- probably would need to look at best practices, but just tracking it

sailsbot:
Thanks for posting, @mikermcneil.  I'm a repo bot-- nice to meet you!

It has been 60 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

rachaelshaw:
@Thomasdezeeuw @snyx1 @mphasize @polastre @aclave1 you can now soft-delete records with `.archive()` - [Docs](https://next.sailsjs.com/documentation/reference/waterline-orm/models/archive)

abalad:
@rachaelshaw The idea of SoftDelete is cool, but we need to create more implementations.
As a BLUEPRINT to be able to perform this operation directly via ENDPOINT

PR #163, opened 2013-10-30
Support for user defined Primary Key & soft deletes

Thomasdezeeuw:
I think this is a better solution that the one I made before for issue #158. Basicly it adds the pk attribute to the schema. Then findOne uses that instead of finding it again. Also normalizing criteria required the schema to be their for the primary key.

Also had to change the test to account for the extra attribute.

A beginning for soft deletes, fixes https://github.com/balderdashy/waterline/issues/134.


harishanchu:
Whether waterline support soft deletes?? because I didn't see soft deletes yet in the docs.

Issue #427, opened 2014-05-14
Denormalization support? Discuss, review or feedback


clarkorz:
When using mongodb, it's very common to use [denormalizated schema](http://en.wikipedia.org/wiki/Denormalization), to make single database query for each request. 

For example, `User` model may have many properties (displayName, avatar, location, address, cellphone, ...), `Blog` model has an `author` attribute to associate with `User` model, but not only with `user.id`, but also `user.displayName`. If **Listing Blogs** only need to display user's name, then `GET /blogs` request only need to query `blogs` collection ONCE without querying `users` collection. Of cause there is cost: data redundancy. But user won't change their `displayName` every day.

Hence I think it will be great that sails could support denormalization.

My teammates were going to replace associations with simple sub-documents, until I worked out certain commits on [waterline](https://github.com/clarkorz/waterline/tree/denormalization), [waterline-schema](https://github.com/clarkorz/waterline-schema/tree/denormalization), and [sails-mongo](https://github.com/clarkorz/sails-mongo/tree/denormalization). 

The idea is to add an `embed` property on association config, it can has an array of attribute names, then the child model (with specific properties ONLY) will be embed to the parent model, and this embedded document can represent `childKey`(or `foreignKey`). Basically, using association api just like before( with foreign key, saved model object, or new model object), then the associated models will be saved and retrieved as sub-document of parent models in mongodb,  instead of just foreign keys.

	// Blog
	module.exports = {
	  attributes: {
		title     : { type: 'string' },
		content   : { type: 'string' },
		author    : { model: 'User', required: true, embed: ['nickname'] }
	  }
	}

    User.create({nickname: 'clark', birthday: '2014-01-01'}).exec(function (err, user) {
      Blog.create({title:'', content: '', author: user}).exec(function (err, blog) {
        console.log(blog);
        // { id: '5373c6bcd9d5d200005c31e9',  title:'', content: '', author: { id: '5373c6bcd9d5d200005c31e8', nickname: 'clark' }, ... }
        // > db.blogs.find():
        // { _id: ObjectId('5373c6bcd9d5d200005c31e9'), author: { _id: ObjectId('5373c6bcd9d5d200005c31e8'), nickname: 'clark' }, ... }
      });
    });

    sails.config.blueprints.populate = false // set `false` to make single query on `find` and `findOne`

arxpoetica:
[...]

PR #428, opened 2014-05-14
Support embeddable association (denormalization)

clarkorz:
With certain commits in `waterline-schema` and `sails-mongo`,
association can be set to embeddable using `embed` keyword. And if so,
the associating model can be fully or partially embedded. This is
denormalization, and only works with `sails-mongo` currently.

https://github.com/balderdashy/waterline/issues/427

### DIFF
--- a/lib/waterline/core/schema.js
+++ b/lib/waterline/core/schema.js
@@ -148,6 +148,11 @@ Schema.prototype.objectAttribute = function(value) {
         attr.size = value[key];
         break;
 
+      // Set schema[attribute].embed
+      case 'embed':
+        attr.embed = value[key];
+        break;
+
       // Handle Belongs To Attributes
       case 'model':
         var type;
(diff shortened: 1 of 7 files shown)


leobudima:
Would be absolutely great if we could push this through!

arxpoetica:
For anyone wondering what this all means, here's some light reading: http://blog.mongodb.org/post/87892923503/6-rules-of-thumb-for-mongodb-schema-design-part-2

elennaro:
+1 for this

particlebanana:
Hey @clarkorz thanks for your time on this. I know it's been sitting here for a while and that's mainly because I felt guilty closing it after the time that went into it. Now it's been sitting long enough that it can't be merged.

I have denormalization on the roadmap. It's something I would like to see a full spec on, especially how deep querying could work but it just doesn't fit into the project right now.

If anyone feels very passionately about this and wants to write up a spec based on the [Contribution Guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) that would be great!

PR #525, opened 2014-07-13
create beforeFind lifecycle callback

aclave1:
I created a beforeFind lifecycle callback to allow a user to always modify a model's find or findOne queries with some criteria. A good use case would be if you're using a soft-delete system and you'd like to tack on an isDeleted:false to any find query on a model. Please feel free to critique the code/suggest changes, I tried to stick to the style that already exists.

### DIFF
--- a/lib/waterline/query/finders/basic.js
+++ b/lib/waterline/query/finders/basic.js
@@ -13,7 +13,8 @@ var usageError = require('../../utils/usageError'),
     waterlineCriteria = require('waterline-criteria'),
     _ = require('lodash'),
     async = require('async'),
-    hasOwnProperty = utils.object.hasOwnProperty;
+    hasOwnProperty = utils.object.hasOwnProperty,
+    callbacks = require('../../utils/callbacksRunner');
 
 module.exports = {
 
@@ -39,6 +40,10 @@ module.exports = {
 
     // Normalize criteria
     criteria = normalize.criteria(criteria);
+
+    this.beforeCallbacks.call(self,criteria.where,function(err){
+      if (err) {return cb(err);}
+    });
 
     // Return Deferred or pass to adapter
     if(typeof cb !== 'function') {
@@ -205,6 +210,10 @@ module.exports = {
       return usageError('Invalid options specified!', usage, cb);
     }
 
+    this.beforeCallbacks.call(self,criteria.where,function(err){
+      if (err) return cb(err);
+    });
+
     // Return Deferred or pass to adapter
     if(typeof cb !== 'function') {
       return new Deferred(this, this.find, criteria);
@@ -351,6 +360,15 @@ module.exports = {
       }
 
     });
+  },
+
+  beforeCallbacks:function(criteria,cb){
+    var self = this;
+    async.series([
+      function(cb){
+        callbacks.beforeFind(self,criteria,cb);
+      }
+    ]);
   },
 
   where: function() {
--- a/lib/waterline/utils/callbacks.js
+++ b/lib/waterline/utils/callbacks.js
@@ -10,5 +10,6 @@ module.exports = [
   'beforeCreate',
   'afterCreate',
   'beforeDestroy',
-  'afterDestroy'
+  'afterDestroy',
+  'beforeFind'
 ];
--- a/lib/waterline/utils/callbacksRunner.js
+++ b/lib/waterline/utils/callbacksRunner.js
@@ -138,3 +138,21 @@ runner.afterDestroy = function(context, values, cb) {
 
   async.eachSeries(context._callbacks.afterDestroy, fn, cb);
 };
+
+/**
+ * Run before Find Callbacks
+ *
+ * @param {Object} context
+ * @param {Object} values
+ * @param {Function} cb
+ * @api public
+ */
+
+runner.beforeFind = function(context, values, cb) {
+
+  var fn = function(item, next) {
+    item(values, next);
+  };
+
+  async.eachSeries(context._callbacks.beforeFind, fn, cb);
+};
--- a/test/unit/callbacks/beforeFind.find.js
+++ b/test/unit/callbacks/beforeFind.find.js
@@ -1,0 +1,155 @@
+var Waterline = require('../../../lib/waterline'),
+    assert = require('assert');
+
+describe('.beforeFind()', function() {
+
+  describe('basic function', function() {
+    var person;
+
+    before(function(done) {
+      var waterline = new Waterline();
+      var Model = Waterline.Collection.extend({
+        identity: 'user',
+        connection: 'foo',
+        attributes: {
+          name: 'string',
+          isDeleted:'boolean'
+        },
+
+        beforeFind: function(values, cb) {
+          values.isDeleted = false;
+          cb();
+        }
+      });
+
+      waterline.loadCollection(Model);
+
+      // Fixture Adapter Def
+      var adapterDef = {
+        find: function(con, col, values, cb) { return cb(null, [values.where]);},
+        findOne:function(conn,col,values,cb){return cb(null,values.where);}
+      };
+
+      var connections = {
+        'foo': {
+          adapter: 'foobar'
+        }
+      };
+
+      waterline.initialize({ adapters: { foobar: adapterDef }, connections: connections }, function(err, colls) {
+        if(err) done(err);
+        person = colls.collections.user;
+        done();
+      });
+    });
+
+    // /**
+    //  * find
+    //  */
+
+    describe('.find()', function() {
+
+      it('should run beforeFind and mutate values', function(done) {
+
+        person.find({name:'billy'}, function(err, users) {
+          assert(!err);
+          assert(users[0].isDeleted === false);
+          done();
+        });
+
+      });
+    });
+
+
+    // /**
+    //  * findOne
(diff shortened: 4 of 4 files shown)


KirillSuhodolov:
very useful improvement
what's current status?

mphasize:
I created a Policy that enables this behavior. Here's the Gist: https://gist.github.com/mphasize/e9ed62f9d139d2152445

aclave1:
I have to agree with

tjwebb:
I'm working on solidifying sails-permissions for this kind of access control behavior: https://github.com/tjwebb/sails-permissions. There's still work to do, but hopefully some will find it to be a good starting point.

randallmeeker:
I was hoping for this feature, not for access control, but for the scrubbing of query params among other reasons. I hate to see it go. 

Another use case would be query limiting where you want to make sure people specify certain params always and set defaults when doing a query.  

What this mostly does, it gives us freedom. By having this (and afterFind), give the ability for the users to implement specific use cases that waterline does not cover, helps keep controllers slim and our code DRY.

Not to mention that is falls in line with the convention of giving lifecycle callbacks to all core methods.

tjwebb:
I liked this feature too. That said, it may be that it could be too easily used for evil.

I don't think it should be forgotten forever. I'm willing to re-open it given a sufficiently convincing argument.

connor4312:
I think that @aclave1 made a good point. I found my way here looking for a way to do soft deletes in waterline. Particlebannana mentioned that they were on the roadmap, but disappointingly that [no longer seems to be the case](https://github.com/balderdashy/waterline/blob/master/ROADMAP.md#roadmap).

randallmeeker:
I'm not sure why this is evil? Or any more or less evil than beforeValidate,beforeCreate,beforeUpdate ?

I'm trying to think of a framework that had callbacks, but did not have ```beforeFind``` and ```afterFind``` and I think sails is the only exception. (cause you know I have used everything, ever . .. probably not)

@tjwebb I'm not sure what a convincing argument would be that has not been made. What is the arguments for beforeValidate,beforeCreate,beforeUpdate?

Most people today in MVC are talking about skinny controllers and dry programming. When we put functionality in the model, we help accomplish these goals. I get that their needs to be a difference between controller logic and model logic, but we are already bluring these lines with the toJSON/toObject methods (as well as all the other callbacks).

A lot of these things can be done in Policies, but that is not ideal. Policies can be dangerous because they can be bypassed easy enough when writing out your model calls in a controller. 

Example: I need to make sure that I include a security credential with every call to my user database to filter by. The check for this security credential is done at the policy level, however I could write a controller where I forgot to use the credential when searching despite being present. A beforeFind callback could check for the existence of this required field for searching, thus forcing me to specify it or returning an error. This goes a long way to providing better security for an app.

In the same manner I can also add checks in the before find to make sure that certain criteria are specified in the query. For instance if in order to limit bloated search results I required that every search require a date range. 

BeforeFind: soft deletes. I know their is a plan in the works, but my preference on the the soft deletes is to use a date field ```deletedAt``` in the same manner that ```updatedAt``` and ```createdAt```. I know this is not the current implementation that is being worked on, but a beforefind callback would allow me to implement this anyway I want. CRY FREEDOM!

BeforeFind: Scrub data for simplified data params. If I know that every search in my user table needs to search Last Name, First Name and Email Address, I can setup my before find to look for a generic ```_q``` field, and if found in a ```beforeFind()```, then expand it to search for all 3 fields and drop the made up ```_q``` field before moving on. I have done this in other apps and it helps keep stuff really dry and eliminate long repetitive code for common but complex queries. 

I could keep going. Currently I find that I'm using creative policies to do a lot of this work, but I don't think they belong in Policies because they are very model specific.

I'll revert to my previous comment 
>What this mostly does, it gives us freedom.

I get that freedom can be dangerous, but I don't think any more than the other callbacks. 

This is also popular subject having been referenced here:

https://github.com/balderdashy/sails/issues/1141
https://github.com/balderdashy/sails/issues/1005
https://github.com/balderdashy/sails/issues/2760
https://github.com/balderdashy/waterline/pull/798
https://github.com/balderdashy/waterline/issues/134

Its also been mentioned in different forms on StackOverflow and the Google Groups and I did not even look in the individual adapter repos.

luislobo:
I still think we need this. There are tons of use cases.

aclave1:
@luislobo can you provide us with some more use cases? I would love for this feature to be included, but I may have to rewrite some of this code as waterline has likely changed enough that merging this would be very hard.

luislobo:
Some of them were already mentioned, by @randallmeeker  and others, but I'll enumerate them:
- Ability to add criteria to all finds, independant of the actual filters a user might select, for example, injecting userId, customerId, ownerId, etc (owners of data)
- Filtering soft deleted objects
- Filtering sitewide data, for example, an international site might show certain data based on their domain and country.
- Adding conditions to user selected filters, that are hidden for them.
- Enforcing date ranges
- Enforcing limits
- The need to always retrieve info from other table/collection that needs to be expanded/used as a condition to the current model. For example: a site that have users, that are assigned to groups, groups can also be assigned to groups, and data "belong" to some of those groups. We have to be able to ALWAYS query and ENFORCE for all finds in the whole application, that for a particular Equipment, it belongs to a user, based on the groups the user is member of, and the group hierarchy.

Some considerations mentioned:
- Having this in a single method, you don't have to repeat it everywhere, as those conditions added are cross site. It helps in making sure that you do it once, and it will be enforced always. I don't want to add deleted=false for all my models and queries!
- Along the same lines: Model specific stuff in models, not in controllers/services

Those are some of the reasons

devinivy:
Bear in mind that this is doable and much more configurable using custom collection methods.  You could alternately extend `find`.  Based upon what everyone is saying, though, maybe a more particular feature for waterline would be default/active criteria.

randallmeeker:
I don't think anyone is saying, "I can't do XXXX because I don't have the beforeFind() callback." I think what is being said is the beforeFind() solves many use cases, features and issues that have been brought up. 

I think plenty of reasons have been given and I understand their are bad use cases out there, but can someone make the case of why implementing a beforeFind (AND afterFind) is the wrong thing for sails to do? How it would affect the framework in a negative fashion? and why its more evil than other model callbacks?

devinivy:
Actually I think it's a reasonable feature!  Was just wondering if there's a solution that solves a particular need more precisely.  It sounded like most of the use-cases had to do with default criteria.

aclave1:
Should I reopen this?

aclave1:
I meant to click comment instead of reopem. Oh well!

dmarcelino:
:+1: I'm good with this as well.

particlebanana:
Alright guys you've convinced me. The community has spoken! I'm not ready for a `0.11` release yet but let's create a branch for it. Then we can keep master clean and available for hot fixes while we flush out any breaking changes we want to go ahead and get in `0.11`. I'll spin that up and then we can re-submit this on that branch.

You guys cool with that?

dmarcelino:
@particlebanana, I'm cool with that. In my opinion our strategy should be:

1. **0.10.x**
  * Fix priority bugs;
  * Get waterline - adapters integration tests to pass: [Build Status](https://travis-ci.org/balderdashy/waterline-adapter-tests)
2. **0.11.0**: new relevant features, suggestions:
  * beforeFind;
  * many-to-many through associations;
  * Custom join table names;
  * Recursive populate? (it's our top request)

devinivy:
Seems reasonable to me!  beforeFind, through associations, and custom join tables are in sight.  I know @particlebanana has made strides on recursive populate.  So we're more-or-less on the right track feature-wise.  I'd be be curious to add composite indexes too.  What about other features for which we already have PRs?

randallmeeker:
If I could offer the suggestion that 0.11.0 also include afterFind as well. Might as well since were doing a beforefind and it would complete our lifecycle callbacks on all model methods. I don't think there is a ready pull request, but I can work on that one. The only argument to be made agaist is that is duplicates some of the work being done with toJson/toObject

https://github.com/balderdashy/sails/issues/1005
https://github.com/balderdashy/waterline/pull/798

aclave1:
@particlebanana thanks! if you want me to implement afterFind as well, let me know! i'll submit another PR with that feature.

particlebanana:
Alright guys, added a [0.11](https://github.com/balderdashy/waterline/tree/0.11) branch. Can we re-submit this against that branch?

connor4312:
I think that baked-in support for dealing with soft deletes (would be preferable to having lifecycle callbacks, for that particular problem. Any chance of those being roadmapped in 0.11?

randallmeeker:
@connor4312 the soft delete is still open here https://github.com/balderdashy/waterline/issues/134

taylorlapeyre:
:+1:

aclave1:
resubmitting against 0.11

aclave1:
new pull request: https://github.com/balderdashy/waterline/pull/902

ufnd:
Any news about this feature?

ChrisWorks:
Yes, what happened to this? Did it get in or not?

abalad:
Yes, what happened to this? Did it get in or not?

silverark:
Any news? I can see there are comments in it for "FUTURE: this is where it would go. It's also in the list of allowed overrides, but just no code.

PR #681, opened 2014-10-08
Added guarding/fill ability to waterline models

connor4312:
Guarding ability as seen in #649, just had to move it to a new branch.

Previously, there wasn't a really _good_ way of updating a model's attributes en masse (e.g. from form input) without manually filtering out attributes you didn't want to have the user be able to set. That doesn't lend itself to very DRY design, and is a pain in the neck!

I've added the ability to denote "guarded" properties of the model, then use a method "fill" to effectively extend the model based on dirty input.

For example, you can have a model like:
```js
{
  attributes: {
    id: {
      type: 'integer',
      autoIncrement: true,
      primaryKey: true,
      unique: true
    },
    email: { type: 'email' }
  },
  guarded: ['id']
}
```

Now, by calling `model.fill(input)`, the user would only be able to change their email, _not_ their ID. Additionally, anything given in the "input" which is not an attribute of the model will be discarded.

### DIFF
--- a/lib/waterline/model/index.js
+++ b/lib/waterline/model/index.js
@@ -37,6 +37,10 @@ module.exports = function(context, mixins) {
       return new defaultMethods.destroy(context, this, cb);
     },
 
+    fill: function(attributes) {
+      defaultMethods.fill(context, this, attributes);
+    },
+
     _defineAssociations: function() {
       new internalMethods.defineAssociations(context, this);
     },
--- a/lib/waterline/model/lib/defaultMethods/fill.js
+++ b/lib/waterline/model/lib/defaultMethods/fill.js
@@ -8,13 +8,14 @@ var _ = require('lodash');
 /**
  * Model.fill()
  *
- * Takes the currently set attributes and updates the database.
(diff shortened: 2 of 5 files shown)


particlebanana:
Hey @dmarcelino @randallmeeker @tjwebb @devinivy what do you guys think of this in `0.11`?

devinivy:
Yes!  0.11 could use it.  Alongside `Model#fill`, `Collection#_model` needs a more intuitive and more public counterpart.  I'd like to write,
```javascript
var post = new Post.model();
post.fill(request.payload)
post.save();
```

dmarcelino:
Definitely `0.11`, this looks like something I'll be using!

particlebanana:
I understand the concept but I think there needs to be some work on this. For example what happens on:

```javascript
User.update({ id: 1 })
.set(input)
.exec(/**/)
```

I like the concept and if someone wants to submit a full specced version of this following the new [Contribution Guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) that would be great!

Issue #683, opened 2014-10-09
Unchanged values unexpectedly marked for update


lwansbrough:
Original issue here: https://github.com/balderdashy/sails/issues/2297

The TL;DR of that issue is that the Sails docs say that when `beforeUpdate` is called, the `valuesToUpdate` argument will contain the values that are being updated (this is what I've interpreted it to mean, anyway). This makes sense and should be the way it works. However, this is only true for updates made through the static `Model.update(...)` method, and not the `model.save()` instance method. In the latter case, `valuesToUpdate` will be all the properties of the model instance, not just the values being updated. Typically the suggestion could just be to use the static method if that's a problem, but this is not possible if you need to update an association. Updating an association currently requires using the model instance.

The solution to this problem therefore, is that we either need static methods for adding associations, or the implementation for updating model instances needs to be changed. I prefer both be done actually, I think there's valid reasons to use both. I've outlined how this might be accomplished in the other issue.

Thanks for your consideration guys.

devinivy:
Important to me too.

particlebanana:
Hey guys sorry I'm late to the party. I think something like #710 would be nice for this. Then no matter what you do when you call `.save` it's only updating things that have changed.

sailsbot:
Thanks for posting, @lwansbrough.  I'm a repo bot-- nice to meet you!

It has been 60 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

PR #710, opened 2014-10-31
Detection of changed properties

connor4312:
Added the ability to check changed properties. This was particularly useful to us when seeing if the password was updated updated on a user's `beforeUpdate` event. With this code, one could nicely check:

```js
    beforeUpdate: function (values, callback) {
        if (this.dirty('password')) {
            bcrypt.hash(/* ... */);
        }
    }
```

Without use of additional metadata or manually hashing when the password is updated.

### DIFF
--- a/lib/waterline/model/index.js
+++ b/lib/waterline/model/index.js
@@ -37,6 +37,14 @@ module.exports = function(context, mixins) {
       return new defaultMethods.destroy(context, this, cb);
     },
 
+    dirty: function(attribute) {
+      return defaultMethods.dirty(context, this, attribute);
+    },
+
+    clean: function(attribute) {
+      defaultMethods.dirty.clean(context, this, attribute);
+    },
+
     _defineAssociations: function() {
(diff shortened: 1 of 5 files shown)


dmarcelino:
For reference this can potentially fix #683.

RWOverdijk:
Looks fine... I could pull these changes and check if it fixes things.

dmarcelino:
That would be great, if you don't mind...

RWOverdijk:
I'll pull this in first thing tomorrow.

RWOverdijk:
@dmarcelino This doesn't fix the issue.

dmarcelino:
@RWOverdijk, I've looked at the PR more deeply and I think it only adds the ability to check if something is clean/dirty. I don't think `.save()` would take it into consideration.

RWOverdijk:
That's what I thought, too. But I promised to check, so I checked.

dmarcelino:
Closing this since PR #1048 continues this work.

PR #717, opened 2014-11-05
Bind model context to lifecycle callback

pierreinglebert:
For example, I want to set a "counter" on my model depending on a criteria.
```javascript
beforeValidate: function(values, cb) {
  this.findOne({
    where: criteria,
    sort: { updatedAt: 'desc' }
  }).exec(function(err, user) {
    values.counter = user ? 0 : user.counter + 1;
    cb();
  });
```
Currently the context passed is the model prototype, why not the model ?
I could access to findOne but I got an error because _transformer is undefined.

### DIFF
--- a/lib/waterline/query/validate.js
+++ b/lib/waterline/query/validate.js
@@ -24,7 +24,7 @@ module.exports = {
       // Run Before Validate Lifecycle Callbacks
       function(cb) {
         var runner = function(item, callback) {
-          item(values, function(err) {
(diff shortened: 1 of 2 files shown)


hardchor:
Agreed. I ran into this a few times now and had to awkwardly work around it by fetching the record (again)

dmarcelino:
I wasn't aware of this, nice one! :clap:  There should be something about this in waterline-docs.

Issue #687, opened 2014-10-11
/lib/waterline/error/WLValidationError.js references non-existant globalId


alarner:
I am getting ambiguous error messages from the waterline validator. For example:

```
Invalid attributes sent to undefined:
 • rgt
   • `undefined` should be a integer (instead of "[object Object]", which is a object)
```

I traced one of the undefined references to /lib/waterline/error/WLValidationError.js line 47 which references self.globalId, but I cannot seem to find where globalId is set. If I replace the reference to self.globalId with:

`self.globalId || self.adapter.identity`

The error message appears to be fixed. I can submit a pull request if you'd like, but I'm not sure if this is the appropriate fix.

devinivy:
Let's crush these!
https://github.com/balderdashy/waterline/search?utf8=%E2%9C%93&q=globalId

dmarcelino:
So is `globalId` meant to be the name of the adapter being used?

Funny enough there was already a PR to address this: #736, merged on Nov 19, 2014.

ericksepulveda:
+1, seems to me that the message should be Invalid attributes sent to modelName, which would be really useful.

sailsbot:
Thanks for posting, @alarner.  I'm a repo bot-- nice to meet you!

It has been 30 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

PR #736, opened 2014-11-19
Update globalId from validation

marsanla:
Allow to validate rules outside from sailsjs


dmarcelino:
For reference this should have fixed #687.

PR #759, opened 2014-12-12
+ [feature] supports the enum/in can use object to transform value

snowyu:
these PR MUST BE applied first:

* https://github.com/balderdashy/anchor/pull/67
* https://github.com/chriso/validator.js/pull/336

see https://github.com/balderdashy/waterline/issues/52

```json
attributes: {
  state: {
    type: 'int',
    enum: {'pending':2, 'approved':3, 'denied':4]
  }
}
```

### DIFF
--- a/test/unit/validations/validation.enum.js
+++ b/test/unit/validations/validation.enum.js
@@ -3,7 +3,7 @@ var Validator = require('../../../lib/waterline/core/validations'),
 
 describe('validations', function() {
 
-  describe('enum', function() {
+  describe('enum string', function() {
     var validator;
 
     before(function() {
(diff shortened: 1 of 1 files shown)


particlebanana:
@snowyu can you refresh me on what this does? What is the purpose of the object notation?

snowyu:
This an unit test only. the real code is https://github.com/balderdashy/anchor/pull/67.
Last year I said  the chriso (validator author) merged partly. so you have to add or enhance the interface( Anchor.match.type() to pass the additional rule information.

and the feature issue is https://github.com/balderdashy/waterline/issues/52.

particlebanana:
Changing the data based on a validation isn't really the goal of the validation system in Waterline. Custom validations or lifecycle callbacks could be used for this currently.

PR #769, opened 2014-12-22
Added missing err checking in create portion of findOrCreate()

ruel:
Whenever create fails in findOrCreate, the err variable is always null. I've added the missing `cb(err)`.


particlebanana:
Thanks!

PR #787, opened 2015-01-08
[feature] Add Multi-Tenant support

Glavin001:
## About

Support [Multitenancy](http://en.wikipedia.org/wiki/Multitenancy) by using *dynamically* loaded Tenant-specific collections and their corresponding connections.

### Example App with Multitenancy

See https://github.com/Glavin001/sails-multitenancy-example

### How It Works

The call to `Collection.tenant` will return a tenant-specific collection. This collection has a scoped `identity` for each `tenant` and the tenant name could be a number of string.
For instance, Collection with identity `test` would becoming `test<tenant:1>` for `Test.tenant(1)`.

The Tenant-specific Collection also has it's own Tenant-specific Connection with an analogous naming convention: `connectionName<tenant:tenantId>`.

These new Tenant-specific collections and connections are loaded dynamically, on-demand, at runtime. This means that you:
1) do not require waiting to load all `X` number of connections for all of your tenants, and
2) do not even have to know at server startup what tenants are available. They could even be loaded from a database upon request, without reloading your server!
 
### Related Issues

See #286 and https://github.com/balderdashy/sails-mongo/issues/130 and https://github.com/balderdashy/sails/issues/1857

### Configure Connection

```javascript
      'my_foo': {
        adapter: 'foo',
        host: 'localhost',
        port: 12345,
        database: 'default_database',

        // === New ===
        isMultiTenant: true, // Enable Multi-Tenancy feature
        defaultTenant: false, // false, <string>, or <number>
        getTenant: function(req, cb) { // For Sailsjs
          return cb(req.params.tenant);
        },
        configForTenant: function(tenantId, config, cb) { // For Waterline
          // Validate Tenant
          tenantId = parseInt(tenantId);
          if (tenantId >= 1 && tenantId < 10) {
            config.database = "tenant-" + tenantId;
            return cb(null, config);
          } else {
            // 
            return cb(new Error("Invalid tenant " + tenantId + "!"));
          }
        }

      }
```

#### Currently Supported

- [x] Callback-style

```javascript
      Collection
        .tenant("1", function(err, TenantCollection) {
          // Now you have a Tenant-specific Collection!
          TenantCollection
            .find({})
            .exec(function(err, results) {
              assert(results.length === 1, 'tenant-1 has 1 record');
              assert(results[0].message === db['tenant-1']['tests'][0].message, 'records are from the tenant-1 database');
              done();
            });
        });
```

- Promise-style, using
  - [x] Operations
    - Used by `find`ers
  - [x] Deferred
    - Used by `count`, `create`, `destroy`, `update`

```javascript
      Collection
        .tenant("1")
        .find({})
        .exec(function(err, result) {
          console.log(err, result);
          done();
        });
```

## Tests: 
See https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/query/dql/tenant.js

- [x] Callback-style API
  - [x] find
  - [x] findOne
  - [x] count
  - [x] create
  - [x] destroy
  - [x] update
- [x] Promise-style API (`Operation`/`Deferred`)
  - [x] find
  - [x] findOne
  - [x] count
  - [x] create
  - [x] destroy
  - [x] update
 
### TODO:

- [x] Use existing Tenant connection, if previously loaded
- [x] Use existing Tenant Collection, if previously loaded
- [x] Remove debugging `console.log`
- Add support for Promise-style API (see above)
- Add support to Sailsjs Blueprints for using this Multi-Tenancy feature
  - Sailsjs would use the `.getTenant(req, callback)` as defined in the configuration
- [x] Improve documentation
- [ ] Joins/associations across multiple multi-tenant Collections
  - Apply the same tenant to both, if either have multi-tenant enabled
- [ ] Allow for the `Adapter` to change between Tenants
  - See https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/connections/index.js#L67 and https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/query/dql/tenant.js#L182

### DIFF
--- a/lib/waterline/collection/index.js
+++ b/lib/waterline/collection/index.js
@@ -54,3 +54,39 @@ inherits(Collection, Query);
 
 // Make Extendable
 Collection.extend = extend;
+
+
+/**
+* Resolve Collection
+*
+* Helper to resolve `tenant` into
+* Tenant-specific Collection, if applicable
+*
+* @param {String} connection name
+* @param {Function} callback(err, resolvedCollection)
+* @return this
+*/
+Collection.prototype._resolveCollection = function(connectionName, cb) {
+    // Get Connection
+    var connectionName = connectionName || this.adapter.dictionary.identity;
+    var connection = this.connections[connectionName];
+    var config = connection.config;
+    // Check if Multi-Tenant enabled
+    if (config.isMultiTenant) {
+        // Is Multi-Tenant connection
+        // Check if there is a Tenant set
+        var tenant = config.tenant;
+        if (tenant) {
+            // Tenant is set
+            // Use Tenant-specific collection
+            // Check if Tenant's connection has been loaded
+            //   console.log('select tenant', config.tenant);
+            this.tenant(tenant, function(err, tenantCollection) {
+                return cb(null, tenantCollection);
+            });
+        }
+    } else {
+        cb(null, this);
+    }
+    return this;
+};
--- a/lib/waterline/collection/index.js
+++ b/lib/waterline/collection/index.js
@@ -64,7 +64,7 @@ Collection.extend = extend;
 *
 * @param {String} connection name
 * @param {Function} callback(err, resolvedCollection)
-* @return this
+* @return {WLCollection} this
 */
 Collection.prototype._resolveCollection = function(connectionName, cb) {
     // Get Connection
@@ -86,7 +86,7 @@ Collection.prototype._resolveCollection = function(connectionName, cb) {
             });
         } else {
             // Error: No tenant set
-            return cb(new Error('Tenant is required to be specified in operation.'), this);
+            cb(new Error('Tenant is required to be specified in operation.'), this);
         }
     } else {
         cb(null, this);
--- a/lib/waterline/collection/index.js
+++ b/lib/waterline/collection/index.js
@@ -82,7 +82,7 @@ Collection.prototype._resolveCollection = function(connectionName, cb) {
             // Check if Tenant's connection has been loaded
             //   console.log('select tenant', config.tenant);
             this.tenant(tenant, function(err, tenantCollection) {
-                return cb(null, tenantCollection);
+                return cb(err, tenantCollection);
             });
         } else {
             // Error: No tenant set
--- a/lib/waterline/collection/index.js
+++ b/lib/waterline/collection/index.js
@@ -84,6 +84,9 @@ Collection.prototype._resolveCollection = function(connectionName, cb) {
             this.tenant(tenant, function(err, tenantCollection) {
                 return cb(null, tenantCollection);
             });
+        } else {
+            // Error: No tenant set
+            return cb(new Error('Tenant is required to be specified in operation.'), this);
         }
     } else {
         cb(null, this);
--- a/lib/waterline/query/dql/tenant.js
+++ b/lib/waterline/query/dql/tenant.js
@@ -9,6 +9,7 @@ var normalize = require('../../utils/normalize');
 var Deferred = require('../deferred');
 var hasOwnProperty = utils.object.hasOwnProperty;
 var CollectionLoader = require('../../collection/loader');
+var extend = require('../../utils/extend');
 
 /**
  * Tenant
@@ -77,39 +78,28 @@ module.exports = function(tenant, cb) {
 
         function createTenantCollection() {
 
+            // console.log('create tenant collection', self instanceof require('../../collection'), self);
+
             var newCollectionIdentity = (baseCollectionIdentity + tenantSuffix).toLowerCase();
 
-            // Take new config and register it's connection
-            // var cols = self.waterline._collections;
-            // console.log('collections', cols);
-            // var col = _.find(cols, function(col) {
-            //     return true;
-            // });
-            // var col = self.collection;
-            // console.log(col);
-            // var colDupModel = _.cloneDeep(self);
-            // var Collection = require('../../collection');
-            //
-            // console.log('connections', self.connections);
-            var colDupModel = self.constructor.extend({
-                // // Set the named connections
-                connections: self.connections,
-                connection: [newConnectionIdentity],
-                //
-                // // Cache reference to the parent
-                // waterline: self.waterline,
-                //
-                // // Default Attributes
-                attributes: self.attributes || {},
-                //
-                tableName: self.tableName,
-                identity: newCollectionIdentity,
-                baseIdentity: baseCollectionIdentity
+            var WaterlineCollection = require('../../collection');
+            var colDupModel = extend.call(WaterlineCollection, {
+                    // Set the named connections
+                    connections: self.connections,
+                    connection: [newConnectionIdentity],
 
+                    // Default Attributes
+                    schema: self.schema,
+                    attributes: self.attributes || {},
+                    tableName: self.tableName,
+                    autoCreatedAt: self.autoCreatedAt,
+                    autoUpdateAt: self.autoUpdateAt,
+                    autoPK: self.autoPK,
+
+                    //
+                    identity: newCollectionIdentity,
+                    baseIdentity: baseCollectionIdentity
             });
-
-            // Change the connection
-            // console.log('colDupModel', colDupModel);
 
             // Load Schema for Tenant's Collection
             self.waterline.schema[newCollectionIdentity] = self.waterline.schema[self.identity];
--- a/lib/waterline/query/dql/tenant.js
+++ b/lib/waterline/query/dql/tenant.js
@@ -14,13 +14,12 @@ var extend = require('../../utils/extend');
 /**
  * Tenant
  *
- * @param {Object} criteria
- * @param {Object} options
+ * @param {Object} tenant
  * @param {Function} callback
- * @return Deferred object if no callback
+ * @return this if no callback
  */
 
-module.exports = function(tenant, cb) {
+Tenant = module.exports = function(tenant, cb) {
   var self = this;
   // console.log('tenant', arguments);
   var usage = utils.capitalize(this.identity) + '.tenant(tenantId,callback)';
--- a/lib/waterline/query/dql/tenant.js
+++ b/lib/waterline/query/dql/tenant.js
@@ -84,6 +84,7 @@ Tenant = module.exports = function(tenant, cb) {
             // There is already an existing TenantCollection
             return cb(null, tenantCollection);
           }
+
           // No existing TenantCollection found
           // Create a new TenantCollection!
           var WaterlineCollection = require('../../collection');
@@ -101,7 +102,8 @@ Tenant = module.exports = function(tenant, cb) {
 
           // Load Collection with new connection
           self.waterline.loadCollection(colDupModel);
-          var loader = new CollectionLoader(colDupModel, self.connections,
+          var loader = new CollectionLoader(colDupModel, self.waterline
+            .connections,
             self.defaults);
           var newCollection = loader.initialize(self.waterline);
 
@@ -116,7 +118,7 @@ Tenant = module.exports = function(tenant, cb) {
         }
 
         // Check if the connection already exists
-        if (self.connections[newConnectionIdentity]) {
+        if (self.waterline.connections[newConnectionIdentity]) {
           // Connection already exists, do not create another
           return createTenantCollection();
         } else {
@@ -135,18 +137,43 @@ Tenant = module.exports = function(tenant, cb) {
           // new Tenant-specific Connection is `this` Collection.
           // However, later on this could support pooling of some kind
           // with multiple Collections for a multi-tenant connection.
-          var tempCollections = {};
-          tempCollections[self.identity] = self;
+          var tempCollections = self.waterline.collections;
           // Grab the schemas used on this connection
           var usedSchemas = {};
-          Object.keys(tempCollections).forEach(function(coll) {
+          Object.keys(tempCollections).forEach(function(collName) {
+            var coll = tempCollections[collName];
+            // Check if collection is a base collection
+            if (coll.baseIdentity) {
+              // Knows it has a base collection
+              // Therefore it is a Tenant-Specific Collection
+              // Only want Base Collections
+              return;
+            }
+            // Check if this collection uses this connection
+            // Normalize connection to array
+            if (Array.isArray(coll.connection)) {
+              // Is array
+              if (coll.connection.indexOf(baseConnectionName) === -1) {
+                // Not found
+                // Collection does not use this Connection
+                return;
+              }
+            } else {
+              // Is not array
+              if (coll.connection !== baseConnectionName) {
+                // Not the same
+                // Collection does not use this Connection
+                return;
+              }
+            }
+            // Continue to add schema
             var identity = coll;
-            if (hasOwnProperty(tempCollections[coll].__proto__,
+            if (hasOwnProperty(coll.__proto__,
                 'tableName')) {
-              identity = tempCollections[coll].__proto__.tableName;
+              identity = coll.__proto__.tableName;
             }
             // Collection is already built
-            usedSchemas[identity] = tempCollections[coll];
+            usedSchemas[identity] = coll;
           });
 
           // Build the connection config
@@ -155,7 +182,8 @@ Tenant = module.exports = function(tenant, cb) {
             _adapter: _.cloneDeep(connection._adapter),
             _collections: []
           };
-          self.connections[newConnectionIdentity] = newConnection;
+          self.waterline.connections[newConnectionIdentity] =
+            newConnection;
 
           newConnection._adapter.registerConnection(config, usedSchemas,
             function(err) {
@@ -169,7 +197,7 @@ Tenant = module.exports = function(tenant, cb) {
 
       });
       return self;
-      
+
     } else {
 
       // Change current tenant on this Collection's Connection's config
--- a/lib/waterline/query/dql/tenant.js
+++ b/lib/waterline/query/dql/tenant.js
@@ -68,6 +68,12 @@ module.exports = function(tenant, cb) {
 
       var configDup = _.cloneDeep(config);
       configDup.tenant = tenant; // Set Tenant
+      // Check if it is the BaseConnection
+      if (connectionName === baseConnectionName) {
+          // Reset the tenant on the connection for the global/main Collection
+          config.tenant = null;
+      }
+      // Get the config for this tenant
       configForTenant.call(self, tenant, configDup, function(err, config) {
         // console.log('configForTenant', err, config);
         if (err) return cb(err);
@@ -79,8 +85,16 @@ module.exports = function(tenant, cb) {
         function createTenantCollection() {
 
             // console.log('create tenant collection', self instanceof require('../../collection'), self);
-
             var newCollectionIdentity = (baseCollectionIdentity + tenantSuffix).toLowerCase();
+
+            // Get TenantCollection, if already exists
+            var tenantCollection = self.waterline.collections[newCollectionIdentity];
+            if (tenantCollection) {
+                // There is already an existing TenantCollection
+                return cb(null, tenantCollection);
+            }
+            // No existing TenantCollection found
+            // Create a new TenantCollection!
 
             var WaterlineCollection = require('../../collection');
             var colDupModel = extend.call(WaterlineCollection, {
@@ -163,10 +177,6 @@ module.exports = function(tenant, cb) {
       });
 
     } else {
-
-      // TODO: Support Promise-Style / Deferred
-      // TODO: Update Deferred to check for Tenant
-      // TODO: Update Operation to check for Tenant
 
       // Change current tenant on this Collection's Connection's config
       config.tenant = tenant;
--- a/lib/waterline/query/dql/tenant.js
+++ b/lib/waterline/query/dql/tenant.js
@@ -1,0 +1,159 @@
+/**
+ * Module Dependencies
+ */
+
+var _ = require('lodash');
+var usageError = require('../../utils/usageError');
+var utils = require('../../utils/helpers');
+var normalize = require('../../utils/normalize');
+var Deferred = require('../deferred');
+var hasOwnProperty = utils.object.hasOwnProperty;
+var CollectionLoader = require('../../collection/loader');
+
+/**
+ * Tenant
+ *
+ * @param {Object} criteria
+ * @param {Object} options
+ * @param {Function} callback
+ * @return Deferred object if no callback
+ */
+
+module.exports = function(tenant, cb) {
+  var self = this;
+  // console.log('tenant', arguments);
+  var usage = utils.capitalize(this.identity) + '.tenant(tenantId,callback)';
+
+  if (arguments.length === 0) {
+    //
+    // Error
+    return usageError('Tenant must be passed.', usage, cb);
+  }
+
+  if (typeof tenant === 'function') {
+    // Tenant is required
+    // Error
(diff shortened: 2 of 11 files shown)


Glavin001:
@particlebanana I am still actively improving the Pull Request right now, however I would appreciate some feedback on the current status of the code. Let me know if there's anything I should change/improve. Thanks in advance!

particlebanana:
@Glavin001 this looks interesting. I love the interface with the `.tenant` chain. Could we do something to not have to acquire the connection first? Currently in adapters each method aquires its own connection already, usually just grabbing one off the pool. If we can tap into that and just look and see if the tenant option was set then it would make it a lot easier. What is the `getTenant` config option for when using sails?

Overall this looks great, good work! I would say lets shoot for something like this and make it where you don't have to manually acquire the tenant connection like in the callback style above.

```javascript
Collection
.tenant(req.session.tenant)
.find({})
.exec(function(err, result) {
    console.log(err, result);
    done();
});
```

Glavin001:
>  Could we do something to not have to acquire the connection first? Currently in adapters each method aquires its own connection already, usually just grabbing one off the pool. If we can tap into that and just look and see if the tenant option was set then it would make it a lot easier. 




Currently I do retrieve the new tenant-specific connection after, when the chain is executed (exec). Could you elaborate on what you mean? I'm new to Sails to I may be missing something. Thanks! 







> What is the getTenant config option for when using sails?




This function / middleware would parse the tenant from the request and could potentially attach it to the session, as you noted.




> make it where you don't have to manually acquire the tenant connection like in the callback style above.




I am still new to Sails so I actually did not even remember we had a req.session. Using the getTenant (change name?) middleware function configured in Sails we could most definitely attach the tenant to the session. Great ideas! 




Thanks for your support. Let me know if there's any more feedback. Much appreciated. :). 


—

lib/waterline/query/dql/tenant.js
```diff
@@ -0,0 +1,194 @@
...
+            // var col = _.find(cols, function(col) {
+            //     return true;
+            // });
+            // var col = self.collection;
+            // console.log(col);
+            // var colDupModel = _.cloneDeep(self);
+            // var Collection = require('../../collection');
+            //
+            // console.log('connections', self.connections);
+            var colDupModel = self.constructor.extend({
```
Glavin001: @particlebanana I am having some problems with this line: `self.constructor.extend`.
I am trying to `extend` (or `clone`) the existing Waterline Collection and change it's `identity` and add `baseIdentity` (identity of base Collection, without tenant suffix).

It works for my Unit Tests, however I am now using https://github.com/balderdashy/sails-mongo and it says that `self.constructor` is a `[Function: bound]` and that errors saying there is no method `extend`. 

```javascript
/Users/glavin/Documents/Streamlyne/api-server/node_modules/sails-mongo/node_modules/mongodb/lib/mongodb/mongo_client.js:475
          throw err
                ^
TypeError: Object function bound() {
        // `Function#bind` spec
        // http://es5.github.io/#x15.3.4.5
        if (partialArgs) {
          // avoid `arguments` object deoptimizations by using `slice` instead
          // of `Array.prototype.slice.call` and not assigning `arguments` to a
          // variable as a ternary expression
          var args = slice(partialArgs);
          push.apply(args, arguments);
        }
        // mimic the constructor's `return` behavior
        // http://es5.github.io/#x13.2.2
        if (this instanceof bound) {
          // ensure `new bound` is an instance of `func`
          var thisBinding = baseCreate(func.prototype),
              result = func.apply(thisBinding, args || arguments);
          return isObject(result) ? result : thisBinding;
        }
        return func.apply(thisArg, args || arguments);
      } has no method 'extend'
  at createTenantCollection (/Users/glavin/Documents/Project Dev/sails/node_modules/waterline/lib/waterline/query/dql/tenant.js:97:48)
  at /Users/glavin/Documents/Project Dev/sails/node_modules/waterline/lib/waterline/query/dql/tenant.js:171:24
  at /Users/glavin/Documents/Streamlyne/api-server/node_modules/sails-mongo/lib/adapter.js:110:9
  at /Users/glavin/Documents/Streamlyne/api-server/node_modules/sails-mongo/lib/connection.js:31:5
  at /Users/glavin/Documents/Streamlyne/api-server/node_modules/sails-mongo/node_modules/mongodb/lib/mongodb/mongo_client.js:472:11
  at process._tickDomainCallback (node.js:463:13)
```

Any help would be greatly appreciated!

randallmeeker:
woot woot +1

Glavin001:
/cc @toddmurphy



—

jfrumar:
This is very exciting and I'll be paying attention to the developments - thanks for your contribution @Glavin001!

I look forwards to replacing our user hook that creates a new set of Models for each tenant with this more efficient (and convenient) approach. One thing we've had to test thoroughly -  under concurrent load Waterline should never access the wrong tenant's DB and introduce crosstalk between tenants. We used  http://jmeter.apache.org/ to test in a staging environment to protect against this.

Glavin001:
> under concurrent load Waterline should never access the wrong tenant's DB and introduce crosstalk between tenants

Absolutely agree. Fortunately, as long as the developer either uses the callback-style, or uses the promise-style and *never stores the promise and uses it later after an asynchronous call* then this will not occur. (See http://stackoverflow.com/a/5177773/3825958 for more info)

### DO NOT

```javascript
var promise1 = Collection
        .tenant("1")
        .find({});

// Async call
// Give control back to Node.js and allow it to run other code first, because this `exec` is called
process.nextTick(function() {
        .exec(function(err, result) {
          // This executed after the code below
          // So this Collection is now a TenantCollection for Tenant="2"
          console.log(err, result);
          done();
        });
});

// Proper usage
Collection
        .tenant("2")
        .find({})
        .exec(function(err, result) {
          // Tenant is 2
          // Same as callback-style, the Colllection is now a `TenantCollection` and will remain correctly scoped to the tenant
          console.log(err, result);
          done();
        });

```


- [ ] Reset the `tenant` to `null` after `exec` is finished
  - This will make sure that the global Collection does not remember the previously set `tenant` after it is executed, and so subsequent requests will have `null` as the `tenant` and will properly error saying that a `tenant is required`.

lib/waterline/query/dql/tenant.js
Glavin001: The above is fixed by https://github.com/Glavin001/waterline/commit/1038dabeec7960638942e798aea6886e5ea2d874

I have tested and confirmed that this works for Sails-mongo now. I would presume it to work on all adapters now!

Glavin001:
Success! Everything is working now. I have added `Collection._resolveCollection` helper function and wrapped [all of the query methods in the `dql/` directory](https://github.com/balderdashy/waterline/tree/bd5152416753d035086825e021e8b11eeb2e0f78/lib/waterline/query/dql). This has majorly improved stability as well as decreased code complexity. I will be removing my changes from the [Operations](https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/query/finders/operations.js) file, now that the lower-level query methods actually handle the resolving of the Tenant Collection, if applicable.

```
$ mocha --recursive --reporter spec test/integration/Collection.multiTenant.js 


  Waterline Collection
    with Multitenancy
      ✓ should have setup correctly 
      using Callback-Style API
        ✓ should find records for specified tenant-1 
        ✓ should find records for specified tenant-2 
        ✓ should count records for specified tenant-1 
        ✓ should findOne record for specified tenant-1 
        ✓ should destroy record for specified tenant-1 
        ✓ should update record for specified tenant-1 
        ✓ should create a record for specified tenant-1 
      using Promise-Style (Fluent Interface) API
        ✓ should error complaining no specified tenant 
        ✓ should error complaining no specified tenant 
        ✓ should find records for specified tenant 
        ✓ should findOne record for specified tenant-1 
        ✓ should count records for specified tenant-1 
        ✓ should destroy record for specified tenant-1 
        ✓ should update record for specified tenant-1 
        ✓ should create a record for specified tenant-1 
        ✓ should find records for multiple tenant requests 
        ✓ should error from improper usage 


  18 passing (58ms)
```

Glavin001:
@particlebanana I'd say everything is complete from a `Waterline` perspective. I will likely need a little help with the Sails part. I'd like to add the custom middleware to every controller action (blueprint and not). The custom middle will use the configured `getTenant(req,cb)` function and get the `tenant` and set `req.session.tenant = tenant`. Could you give me some pointers on this one? I want to make sure I follow your/Waterline/Sails standard contribution practices, and that everyone will be happy to merge and enjoy. Thanks!

Glavin001:
Once this is merged in we can get Sails up to snuff :smiley:. 
See PR: https://github.com/balderdashy/sails/pull/2553

particlebanana:
@Glavin001 I'm trying to find some time to set aside and go through this whole thing and get it merged. This is a big one and I just want to make sure everything is going to be seamless. I should have it in before the weekend if everything looks good.

Glavin001:
Good things take time. I appreciate you setting time aside to review this. Let me know if there's anything I can do.

lib/waterline/query/dql/tenant.js
```diff
@@ -0,0 +1,193 @@
...
+            var tenantCollection = self.waterline.collections[newCollectionIdentity];
+            if (tenantCollection) {
+                // There is already an existing TenantCollection
+                return cb(null, tenantCollection);
+            }
+            // No existing TenantCollection found
+            // Create a new TenantCollection!
+
+            var WaterlineCollection = require('../../collection');
+            var colDupModel = extend.call(WaterlineCollection, {
```
Glavin001: @particlebanana I think this is the only part that I am questionable on.

Objective: Create an extended copy of the base Collection, that uses the `newConnectionIdentity` as the `connection`.

Possible changes: Right now I believe I am ignoring custom class methods. Use [_.merge](https://lodash.com/docs#merge) to merge `{connection: [newConnectionIdentity]}` on top of the `base Collection` into a new object, thus creating a proper clone. I found that [_.cloneDeep](https://lodash.com/docs#cloneDeep) did not give me the desired result.

I will try to look at this today / tomorrow in preparation for your reviewing this weekend. I think this is the last thing I need to clean up. Let me know if you have any other ideas. Thanks.

Glavin001: Fixed with https://github.com/Glavin001/waterline/commit/afa2901e43cb1574f0bbe79cec9c1f9a51098c7c#diff-6ccdf090b069bfba17ecbc7d1ef471f2R89

That's everything I can think of.

lib/waterline/collection/index.js
```diff
@@ -54,3 +54,42 @@ inherits(Collection, Query);
...
+        // Is Multi-Tenant connection
+        // Check if there is a Tenant set
+        var tenant = config.tenant;
+        if (tenant) {
+            // Tenant is set
+            // Use Tenant-specific collection
+            // Check if Tenant's connection has been loaded
+            //   console.log('select tenant', config.tenant);
+            this.tenant(tenant, function(err, tenantCollection) {
+                return cb(null, tenantCollection);
```
Glavin001:
```js
return cb(null, tenantCollection);
```

Should be

```js
return cb(err, tenantCollection);
```

Note the missing `err` when calling the callback.

Special thanks to @chipgata for being a tester and showing some strange and unexpected errors. Fixing this should clear things up.

Glavin001: Fixed with https://github.com/balderdashy/waterline/commit/f7c31ea8d4fb019e58cf2c864dc32e4b7689bfa8

particlebanana:
Hey @Glavin001 can you undo all the formatting stuff? It makes it really hard to see and follow what has changed when there is a bunch of lines like: https://github.com/balderdashy/waterline/pull/787/files#diff-4bba40afded661fbbdf387f7d1e95dc3R135

particlebanana:
Hey @Glavin001 can you undo all the formatting stuff? It makes it really hard to see and follow what has changed when there is a bunch of lines like: https://github.com/balderdashy/waterline/pull/787/files#diff-4bba40afded661fbbdf387f7d1e95dc3R135

Glavin001:
Yeah I accidentally hit my beautifer while developing and it cleaned all of the formatting. I can imagine it is making it difficult to review. I'll try and find time to revert the formatting back to what it was before and make this a little clearer.

particlebanana:
Also it looks like there will be a memory leak because we are never clearing out the instantiated connections/collections that the tenant generates. If you have 100 tenants you will have 100*n where n is the number of models. 

I think a clearer approach is to look to the `spawnConnection` method on the adapters. 

An ideal solution would be to read the tenant config from the query criteria in deferred and pass that straight through to the adapter's `spawnConnection` method that each adapter method is already wrapped in.

```
User.find({ id: 1 }, { _tenant: { ////// } })
```

Would get read by the postgresql adapter for example and create a one off connection, run the query and close the connection. The waterline collections were never meant to be instantiated dynamically and have a lot of baggage.

Glavin001:
> Also it looks like there will be a memory leak because we are never clearing out the instantiated connections/collections that the tenant generates. If you have 100 tenants you will have 100*n where n is the number of models.

This is *not* a memory leak, it is an intended feature. I do not want to have to recreate each Tenant-Collection and Connection for each new Tenant request. Ideally a connection could be switched to use another database/etc and reused, however that is not the case currently. Even now, when my Sails server starts it takes a second or so just for instantiating the connections (I added some logging because I was curious why it was so slow), and adding this time to *each request* is an unacceptable performance hit. If you have any recommendations for performance improvements please let me know.

> I think a clearer approach is to look to the spawnConnection method on the adapters.

Please post a link to where this `spawnConnection` function is (Waterline, each adapter, etc?). Is it
https://github.com/akonwi/sails-mongo/blob/1ccb7f580c411d8d49ad1829814951fe9b5a2fd8/lib/adapter.js#L441-L459 ? I would really appreciate a cleaner approach for implementing this.

> An ideal solution would be to read the tenant config from the query criteria in deferred and pass that straight through to the adapter's spawnConnection method that each adapter method is already wrapped in.
> 
> ```
> User.find({ id: 1 }, { _tenant: { ////// } })
> ```
> Would get read by the postgresql adapter for example and create a one off connection, run the query and close the connection. 

I do not like this. I prefer a Tenant-specific Collection where the query does not contain any tenant information. 
The goal was that for Sails, requests would have the tenant information parsed and the Model/Collection that the Controller would have access to would be scoped to the Tenant. I could take a Sails application that was previously only single-tenant, and configure the connection and have it automatically work for multi-tenants. This has been accomplished:  https://github.com/balderdashy/sails/pull/2553
I would love to hear feedback from other users what they prefer:

```javascript
User.find({ id: 1 }, { _tenant: { ////// } })
```

or

```javascript
// Multitenancy has been handled automatically by Sails and Waterline
// `User` now represents a Tenant-Specific Collection/Model
User.find({ id: 1 })
```

I know I prefer the latter.

> The waterline collections were never meant to be instantiated dynamically and have a lot of baggage.

That's for sure :stuck_out_tongue_winking_eye:. It was tough to make that work and I would love a cleaner solution, however the current Waterline does not seem to be designed for this and a cleaner solution was not apparent to me. Hopefully maintainers of Sails and Waterline could help clean this up, if possible.

I know between this and the [Sails Pull Request](https://github.com/balderdashy/sails/pull/2553), I have accomplished exactly what I set out to do and the usage and experience is what I want. Any performance improvements and cleaner solutions for accomplishing this are welcome.

Thank you for your feedback and please let me know of any other questions / concerns you have. I would really like to get official support for this (merged) with the Sails and Waterline community and provide developers with a solution to what appears to me to be a major lacking feature, and a big problem.

particlebanana:
The `spawnConnection` method is in each adapter and is responsible for either creating a connection or grabbing a connection from the pool. Each adapter method uses this to get a live connection it can use to run a query. You can see an example in the mongo adapter here: https://github.com/akonwi/sails-mongo/blob/1ccb7f580c411d8d49ad1829814951fe9b5a2fd8/lib/adapter.js#L231-L233

Creating a new connection here should be fast, it shouldn't be a bottleneck. It would certainly be more performant than keeping hundreds of dynamic collections in memory I would think. You could even create connection pools for each tenant which would eliminate the connection overhead. The reason it's slow on lift is due to building up the collection schema, running describe queries, doing migrations if needed, etc. Actually creating a connection using the db client should be fast.

Maybe I'm misunderstanding but it wouldn't be:

```javascript
User.find({ id: 1 })
```

it would be:

```javascript
User.tenant("1").find({ id: 1 })
```

Which behind the scenes would create a new waterline collection for the User with the configured tenant connection and then use that to fulfill the query. What I'm proposing is in the deferred object it transforms that to `User.find({ id: 1 }, { _tenant: { ... } })` when passing to the adapter and let the adapter handle the dynamic connection. The userland usage is the same but it doesn't require the overhead of the dynamic collections.

Glavin001:
> The reason it's slow on lift is due to building up the collection schema, running describe queries, doing migrations if needed, etc. Actually creating a connection using the db client should be fast.

Excellent. I'll look into using that instead of [registerConnection](https://github.com/balderdashy/sails-mongo/blob/master/lib/adapter.js#L81). Thanks for the recommendation.

> Maybe I'm misunderstanding but it wouldn't be:
> ```javascript
> User.find({ id: 1 })
> ```
> it would be:
> ```javascript
> User.tenant("1").find({ id: 1 })
> ```
> Which behind the scenes would create a new waterline collection for the User with the configured tenant connection and then use that to fulfill the query. What I'm proposing is in the deferred object it transforms that to `User.find({ id: 1 }, { _tenant: { ... } })` when passing to the adapter and let the adapter handle the dynamic connection. The userland usage is the same but it doesn't require the overhead of the dynamic collections.

Sounds good, however (correct me if I am wrong) this would not allow for the following usage case:

### Calling the query methods with a callback directly

This skips the usage of `deferred` entirely.

```javascript
User.tenant("1").find({ id: 1 }, function(err, result) { });
// This actually uses `Operation` instead of `Deferred`
// See https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/basic.js#L63
```

So long as developers do not use it in this way, this is not a problem, however I would say is reasonable usage (callback-style versus promise-style, developers choice). Handling this in the query methods also means the `Deferred` and `Operation` does not have to deal with something that is Collection specific and can focus on it's core purpose of deferring execution.

I moved the bulk of the code into https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/collection/index.js#L69 and https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/query/dql/tenant.js so it's very quick and easy to scope a Collection by tenant. See https://github.com/Glavin001/waterline/blob/multi-tenant/lib/waterline/query/dql/count.js#L55

> passing to the adapter and let the adapter handle the dynamic connection

So this could be accomplished in Waterline, somewhere in https://github.com/balderdashy/waterline/tree/master/lib/waterline/adapter ? That'd be great to 1) use `spawnConnection` and be even faster and 2) not have to create Tenant-specific Collections in memory and only have the Connections for the current request be scoped to the Tenant.

I do not see where [`spawnConnection`](https://github.com/akonwi/sails-mongo/blob/1ccb7f580c411d8d49ad1829814951fe9b5a2fd8/lib/adapter.js#L441) is exported and therefore giving me access to use it from Waterline. Could you explain how I could call it from Waterline? It would be great to increase the performance of this feature. Thanks!

Glavin001:
We could forfeit the ability to have reusable, scoped Collections:

```javascript
User.tenant("1", function(UserTenant) {
// UserTenant is scoped and reusable
UserTenant.find().exec(); // Use once
UserTenant.create().exec() // Use again
// etc
});
```

And instead of having the `.tenant` API on the `Collection`, we could put it on `Operation` or `Deferred` such that:

```javascript
User.find().tenant(req.session.tenant).exec();
```

And this would use [`spawnConnection`](https://github.com/akonwi/sails-mongo/blob/1ccb7f580c411d8d49ad1829814951fe9b5a2fd8/lib/adapter.js#L441) to generate a Tenant-specific connection when the request is being created and executed, thus eliminating the need for Tenant-specific Collections and Connections to remain in memory and still allowing for dynamic Tenants.

Thoughts @particlebanana?

The only thing I am missing is as stated above:
> I do not see where [`spawnConnection`](https://github.com/akonwi/sails-mongo/blob/1ccb7f580c411d8d49ad1829814951fe9b5a2fd8/lib/adapter.js#L441) is exported and therefore giving me access to use it from Waterline. Could you explain how I could call it from Waterline? It would be great to increase the performance of this feature. Thanks!

mojotaker:
@Glavin001, @particlebanana , Hello there, @Glavin001  I fully support your idea Goals, I pretty much  Did this on  node + Express app   http://stackoverflow.com/a/24913867/343338

Connection pulling is very important, (saving established connections). It is not memory leak, thats the way we want it to be so that if a tenant that has connected before wants to make future/ subsequent request, sails (node) does not need to go through the reconnection dance.

Im very new to sailsJs and this is the one reason that has prevented me from using it full time, as I am not skilled or knowledgeable enough to implement it myself.

particlebanana:
@Glavin001 sorry for the delay here. I haven't had a chance to look into the feasibility of doing the `spawnConnection` part yet. The `spawnConnection` is not available from Waterline, it would be up to the adapter to implement it. This shouldn't be much work though as most of the adapters that would benefit from multi-tenant already use this method. Lets start with mongo and see what we can do.

What Waterline would need to do is to make sure that the tenant config is passed to the query method. So in the above example `User.find().tenant(req.session.tenant).exec();` the `find` method in the mongo adapter would get the correct connection config for the tenant and pass it to the `spawnConnection` method. If a pool exist for the tenant connection then use it otherwise create a pool for it and pass a connection back.

@mojotaker it should absolutely keep an open pool of connections so that it's not constantly creating new ones. @jfrumar has some insights on the creating multiple models approach he posted in https://github.com/balderdashy/sails-mongo/issues/130#issuecomment-47439453. I think if we can figure out the connection pooling without instantiating new models we can make it so you can use a multi-tenant architecture and still be able to easily scale out horizontally.

Glavin001:
@particlebanana so this would be opt-in for each Adapter to support Multitenancy, by exposing the `spawnConnection` API. I was hoping for a Waterline solution, that would not require this, and while that has been achieved with this above Pull Request , I definitely see your point and the advantages of handling it closer to the Connection instead of Collection.

Unfortunately, I have used up my time I had allocated for this feature. However, I hope I can make some time coming up in these next few weeks and potentially submit another Pull Request.
Please detail any other knowledge / advice that you have on this matter, before I begin developing this feature again, so that another Pull Request will not be developed needlessly. Thank you.

If anyone else is interested in contributing to this new implementation of Multitenancy for Waterline, please let me know. Let's hope we can finally bring this critical feature to the Sails and Waterline community!

t3rminus:
I'd be interested in helping with the development of this, if it's still needed. I somehow missed this pull request, and developed my own (much poorer) solution in the meantime, at the ORM-level in Sails.

Let me know what I can do to help get this pulled!

Glavin001:
@t3rminus: Unfortunately, at the moment I do not see a time in the near future where I will be able to contribute and develop this new version of multitenancy support, as @particlebanana and I have described and discussed above.
This Pull Request accomplishes precisely what I intended and needed at the time, however as noted above there may be an even better way of implementing this. This information came too late. I wanted to make this work at the Waterline level and not require any changes to adapters, and this is accomplished with the pull request above. It should still work although it is of course not being maintained and as you can see there are always merge conflicts created since time has passed. See https://github.com/Glavin001/sails-multitenancy-example for a demo.

My company is no longer developing (postponed) our server with Sails and I am now working on other projects so it may be a few months before I will be back and revisiting this issue and Sails server development.
I hope that when the time comes that I do revisit development of our server, that Sails supports some form of multitenancy and that I will not need to look for another framework.

Issue #788, opened 2015-01-08
Associations named 'type' break other columns


bradshja:
Creating an association named 'type' (maybe only belongsTo associations) causes other columns to misbehave by including the association's data.  As an example, I have models like this:

Account.js:
```
module.exports = {
    ...
    attributes: {
        account_id: {
            type: 'integer',
            autoIncrement: true,
            primaryKey: true
        },
        type: {
            model: 'accounttype',
            columnName: 'type_id'
        },
       devices: {
            model: 'device',
            via: 'account'
       }
    }
};
```
AccountType.js:
```
module.exports = {
    ...
    attributes: {
        type_id: {
            type: 'integer',
            autoIncrement: true,
            primaryKey: true
        },
        type: {
           type: 'string'
        }
    }
};
```
Device.js:
```
module.exports = {
    ...
    attributes: {
        device_id: {
            type: 'integer',
            autoIncrement: true,
            primaryKey: true
        },
        account: {
            model: 'account',
            columnName: 'account_id'
        },
       mac_address: {
            type: 'string'
       }
    }
};
```

Given those models, if I retrieve an account with the API, I expect to get something like this:
```
{
  "devices": [ 
     {
      "device_id": 561665222302,
      "mac_address": "00:02:02:4f:b1:59",
      "account_id": 105149
    }
  ],
  "type": {
    "type_id": 8,
    "type": "Employee"
  },
  "status": {
    "status_id": 1,
    "status": "Active"
  },
  "account_id": 105149,
  "display_name": "Jordan Bradshaw"
}
```
but I get this instead:
```
{
  "devices": [],
  "type": {
    "type_id": 8,
    "type": "Employee"
  },
  "status": {
    "status_id": 1,
    "status": "Active"
  },
  "account_id": [
    {
      "device_id": 561665222302,
      "mac_address": "00:02:02:4f:b1:59",
      "account_id": 105149
    }
  ],
  "display_name": "Jordan Bradshaw"
}
```
As you can see, the devices association is populating the account_id property, which is supposed to just be a numerical column as the primary key of the accounts collection.

After much investigation, I've discovered that this is caused by the Transformation.serialize function being applied to the collection's schema in waterline/core/index.js on line 104.  The serialization function sees the type column and recursively replaces all instances of 'type' with 'type_id.'  I'm not sure how this causes the wackiness above, but removing the serialization line fixes the problem, at the possible expense of causing problems elsewhere that I haven't seen.

[...]

PR #789, opened 2015-01-09
Allow Associations to Use 'type' as a Name

bradshja:
This introduces a small change to core's Transformation.serialize method that allows the depth of serialization to be limited (no limit is the default).  In turn, this allows a fix to #788, which will allow models to use 'type' as an association name.  Previously, if you tried to do this it would overwrite any column definition types with the type association's columnName.

### DIFF
--- a/lib/waterline/core/index.js
+++ b/lib/waterline/core/index.js
@@ -101,7 +101,9 @@ Core._initialize = function(options) {
   this._transformer = new Transformer(schemaAttributes, this.waterline.schema);
 
   // Transform Schema
-  this._schema.schema = this._transformer.serialize(this._schema.schema);
+  // This limits the serialization to one level because it could inadvertently replace important properties
+  // if there are conflicting column names (such as type)
+  this._schema.schema = this._transformer.serialize(this._schema.schema, 1);
 
   // Build up a dictionary of which methods run on which connection
   this.adapterDictionary = new Dictionary(_.cloneDeep(this.connections), this.connection);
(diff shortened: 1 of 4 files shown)


devinivy:
I do not understand the PR at first glance, but if it's solid I think it would be nice to have in 0.11.  @particlebanana @tjwebb @dmarcelino any thoughts?

dmarcelino:
@devinivy, can't say I understand it either... I'll trust @particlebanana on this one and I would like to see some automated tests confirming the behaviour change as this looks like a fairly deep core change.

BTW, I've hit the pr tests re-run button and it broke with an awkward error: [Object.defineProperty(EventsModule, "usingDomains", {](https://travis-ci.org/balderdashy/waterline/jobs/50687153#L174). Maybe we need to merge master into bradshja:master.

particlebanana:
It's not that complicated. We have a concept of a **transformer** in Waterline that loops through an object and changes the keys. It's used when you specify a columnName on an attribute. Waterline will transform any keys to the columnName when sending data to the adapter and when it gets data back from an adapter it will run through and make them the attribute name.

It makes it so adapters don't need to worry about implementing this stuff themselves. The max-depth is good to have anytime we are doing recursive parsing like this but I don't think it solves this issue and would stop associated attributes from be transformed.

The actual issue probably isn't in the transformer. I suspect it's somewhere in waterline-schema when it parses a model definition and decorates it into something that we can send down to an adapter. We just need to add a check in there so it handles properties like this.

Closing and moving to https://github.com/balderdashy/waterline-schema/issues/22 if we find that the issue lies here then we can re-open it. I linked back to this issue in the other library.

PR #795, opened 2015-01-13
Update: query/aggregate.js

imevs:
createEach method: changing async.each to async.eachSeries. 

This allows to insert records in the same order as we expect. Also a lot of integraion tests from https://github.com/balderdashy/waterline-adapter-tests expect this behaviour.

### DIFF
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -65,7 +65,7 @@ module.exports = {
       });
     }
 
-    async.each(valuesList, create, function(err) {
+    async.eachSeries(valuesList, create, function(err) {
       if(err) return cb(err);
       cb(null, records);
     });


particlebanana:
@devinivy @dmarcelino now that we updated the adapter tests I think it's good to merge this if it makes sense. I know most of the tests expect this to be true and it's simple enough.

devinivy:
Re-ran Travis– looks good!

tjwebb:
Thanks @devinivy

particlebanana:
Awesome thanks @devinivy!

brandonsimpson:
Thanks @devinivy it works as expected now.

dmarcelino:
Hey guys (@imevs, @particlebanana, @devinivy, @tjwebb and @brandonsimpson) I would like to re-open this for discussion due to performance concerns. Please take a look at: #980.

particlebanana:
Yeah I'm down. Ideally we check if the adapter defines their own `createEach` method and use that and just use the `.map` as a fallback. This way we could get this to a single insert query in pg and mysql.

PR #814, opened 2015-01-28
update helper.js due to many-to-many associations fails on update

restricted:
When to be updated object contains many-to-many associations and its empty array then waterline helper fails with:
return hop.call(obj, prop);
TypeError: Cannot convert null to object at hasOwnProperty (native)


particlebanana:
hey @restricted can you change this back https://github.com/restricted/waterline/commit/24cad5a864e950bbfc2e051c553a547b1c34ffc1

restricted:
Ок.

particlebanana:
Hey @restricted looking at this now, not quite sure what the problem is. Can you expand on the description some more with code examples?

devinivy:
Perhaps this is related to the somewhat elusive #908.

dmarcelino:
@restricted, can you please add a failing test for this?

particlebanana:
There's not enough information here explaining what the issue is or a failing test showing something isn't working.

PR #815, opened 2015-01-28
Fixes #812, moving destroy of related objects only if matching conditions

### LINKED ISSUE #812: Waterline truncated a join table - this is a MAJOR issue
I had an issue on a production server where all data from a join table was dropped. After days of investigation, I found out that waterline generated a delete statement without any where clause on that table.

Repro steps:
* Create a Model with a many to many collection of elements. eg: Team <--> Player
* Insert Team T1 with Players P1 and P2. Insert Team T2 with Players P3 and P4. This will generate two rows in the Team table, 4 rows in the Player table and 4 row in the Team_Player join table.
* Issue two delete statements for team T1 *at the same time*. This is important and is probable with HTTP double post.
* Waterline will generate the following (extract from my postgres log):
```
2015-01-27 21:57:02 EST LOG:  execute <unnamed>: DELETE  FROM "team" AS "team"  WHERE "team"."id" = $1  RETURNING *
2015-01-27 21:57:02 EST DETAIL:  parameters: $1 = '12'
2015-01-27 21:57:02 EST LOG:  execute <unnamed>: DELETE  FROM "team" AS "team"  WHERE "team"."id" = $1  RETURNING *
2015-01-27 21:57:02 EST DETAIL:  parameters: $1 = '12'
2015-01-27 21:57:02 EST LOG:  execute <unnamed>: DELETE  FROM "team_players__user_teams" AS "team_players__user_teams"  WHERE "team_players__user_teams"."team_players" IN ($1)  RETURNING *
2015-01-27 21:57:02 EST DETAIL:  parameters: $1 = '12'
2015-01-27 21:57:02 EST LOG:  statement: DELETE  FROM "team_players__user_teams" AS "team_players__user_teams"   RETURNING *
```

See the last delete statement that does not have a where clause. This will in fact truncate the table thus removing players from all games.

I traced the code and the following change created this issue: https://github.com/balderdashy/waterline/commit/388473fd207d0be7f15a5e42519a6ecadeea6cc7

I tried commenting the if and while I get an exception because the second delete does not return any result (```var mappedValues = result.map(function(vals) { return vals[pk]; });``` will not contain anything), I don't get any dataloss. Do I need to say that it is better to have an error because of concurrency instead of a silent truncation of all data?

luislobo:
I'm not sure if it covers all cases, but this fixes the issue about deleting related objects when removing a model with an invalid condition

### DIFF
--- a/lib/waterline/query/dql/destroy.js
+++ b/lib/waterline/query/dql/destroy.js
@@ -93,9 +93,11 @@ module.exports = function(criteria, cb) {
 
         if(mappedValues.length > 0) {
           criteria[refKey] = mappedValues;
+          collection.destroy(criteria).exec(next);
+        } else {
+          return next();
         }
 
-        collection.destroy(criteria).exec(next);
       }
 
       async.each(relations, destroyJoinTableRecords, function(err) {


kevinburke:
maybe would be good to add a test that exposes the incorrect behavior?

particlebanana:
@luislobo thanks! I'll add some tests for this and get it published asap.

particlebanana:
Pushed out 0.10.17 which should fix this and added an integration test in wl adapter tests

devinivy:
Thanks, all!

PR #818, opened 2015-01-29
Remove toJSON function from objects being returned from toJSON, to preve...

cha0s:
...nt infinite recursion with msgpack or other recursive object transformation tools

Trying to serialize a model with something like msgpack.encode will kill the application with infinite recursion because it will do something like:

```js
function encode(data) {
  ...

  if ('function' === typeof data.toJSON) {
    encode(data.toJSON());
  }
}
```

but waterline models' default toJSON function returns an object which also has a toJSON function. This PR removes the toJSON method from objects returned from toJSON (and refactors the default toJSON function to DRY it up a bit).


particlebanana:
Thanks @cha0s

PR #820, opened 2015-01-29
Disallow destroy to wipe junction table records.

### LINKED ISSUE #799: Calling destroy on a model using other parameter than id wipes out all relations
I have a model called Taxonmy with a custom ID. It has the following attributes

````javascript
attributes: {
	id : { type: 'string',primaryKey:true, required: true,unique:true },
	runningImport: { type: 'boolean',defaultsTo:false,internal:true,protected:true},
	term: { type: 'string', required: true },
	icon: { type: 'string'},
	requireSubscription: { type: 'boolean'}, 
	synonyms: { type: 'string'},
	description: { type: 'string'}, 
	builtBy : { type: 'string'},
	journalEntries : {
		collection : 'journalEntry',
		via : 'taxonomy'
	},
	parent: { model: 'Taxonomy'}, 
	children: {
		collection:'Taxonomy',
		via : 'parent'
	}
}
```

Calling the following
``` javascript
Taxonomy.destroy({runningImport:false}) //wipes all relations to journalEntries
```
wipes out all relations in the journalEntries many-to-many relations table (```journalentry_taxonomy__taxonomy_journalentries```) even for unaffected records

The only way I found for the relations to be maintanted is to delete record by record, by providing a specific ID on destroy (even fails with an array of IDS). The following works properly.

```javascript
Taxonomy.destroy({id:'expositions'})  //this works as expected
```

This seems unsuitable for many situation where you are destroying entities based on something else than it's id. It also fails if you provide a second criteria to the query. ex.
```javascript
Taxonomy.destroy({id:'expositions',runningImport:false})  //wipes out all relations 
```

devinivy:
Potentially closes #799, #812, #815 – please try this out, as this is a critical area of code.
Also, this should require a new test case.

### DIFF
--- a/lib/waterline/query/dql/destroy.js
+++ b/lib/waterline/query/dql/destroy.js
@@ -87,17 +87,18 @@ module.exports = function(criteria, cb) {
         // If no refKey return, this could leave orphaned join table values but it's better
         // than crashing.
         if(!refKey) return next();
-
+        
+        // An array of ids
         var mappedValues = result.map(function(vals) { return vals[pk]; });
+        
+        // Require some id condition for destroying records, otherwise could drop all records in table.
+        if (mappedValues.length == 0) return next();
+        
+        // At this point refKey and mappedValues are guaranteed to exist.
         var criteria = {};
+        criteria[refKey] = mappedValues;
 
-        if(mappedValues.length > 0) {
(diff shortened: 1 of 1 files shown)


devinivy:
Sorry– didn't see that someone had already submitted a PR!  Closed!

PR #827, opened 2015-02-06
Allowing types to be defined in each adapter

shekhei:
I do think this is quite important, there are cases where supported types might be a bit different with each adapter.

With this change, this will allow each adapter to extend and create their own supported types.

Although this might mean that it becomes more difficult to have models that supports all sorts of different adapters, the chances of the models changing adapter is rather low, this change can make waterline more extensive :)

### DIFF
--- a/lib/waterline/adapter/types.js
+++ b/lib/waterline/adapter/types.js
@@ -1,0 +1,6 @@
+module.exports = {
+  supportsType: function(type) {
+    var adapter = this.connections[connName]._adapter;
+    return adapter.supportsType && adapter.supportsType(type);
+  }
+};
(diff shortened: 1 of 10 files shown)


lib/waterline/adapter/types.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  supportsType: function(type) {
+    var adapter = this.connections[connName]._adapter;
+    return adapter.supportsType && adapter.supportsType(type);
```
tjwebb: So, won't this return `false` for adapters which haven't yet implemented this new method?

shekhei: It would really depend on whether "default" types are definitely supported. If they are definitely supported, I suppose we can add an extra condition in front. What do you think?

shekhei:
The branch has been dirtied by dirty from another pull request

Issue #308, opened 2014-03-09
[FEATURE REQUEST] Recursively populate


pixelfreak:
I'm on Sails 0.10. I have a Model which has a collection of Model which in turn has more Model. I know how to populate the first child but how do I recursively populate the rest of the children?

urielaero:
+1

luolongzhi:
+1

mydearxym:
+1

jaumard:
+1

rakeshmenon:
+1

Thelmos:
Definitely a must.

Batornator:
+1

Any news on this feature? Would be ideal for a current project. Are there plans to switch to knex or will there be native support for a recursive populate?

mattblang:
+1

mattblang:
+1
Any news about this?

liratanak:
+1

jeromewir:
+1

Definitely need this feature !

lghamie:
+1 Totally need this.

mkalish:
+1

felipecm:
+1

besarthoxhaj:
@pixelfreak I got the same issue `res.json(theModel)` and `JSON.stringify(theModel)` won't show up the properties populated.

How did you fixed it?

besarthoxhaj:
Ok this is how I did it:

```js
Posts
.findOne(2)
.populate('comments')
.exec(function (err, itemPost){
	// some error handling
	async.map(itemPost.comments,
		function (comment, callback){
			Comments
			.findOne(comment.id)
			.populateAll()
			.exec(function (errComments, commentItem){
				if(errComments){return callback(errComments)};
				callback(null, commentItem);
			});
		},
		function (errFromIterator, results){
			if(errFromIterator){res.serverError()};

			var itemPostToJSON = itemPost.toJSON();

			itemPostToJSON.comments = results;

			var clone = _.clone(itemPostToJSON);
			
			res.send(clone);
		}
	)
});
```

I'm sure there are better ways, so don't be shy and post them:)

felipecm:
It's ugly.. but works with two levels and only two database calls

```javascript
    async.waterfall([
      function(callback) {
        Groups
          .find({parent:parent})
          .populate('nodes')
          .exec(function(err,data){
            callback(err, data);
          });
      },
      function(item, callback) {

        var data = _.flattenDeep(_.map(item,function(obj){
[...]

PR #829, opened 2015-02-07
Decreased processing time by making queries which involve population more efficient

nbiton:
This PR includes two simple changes which vastly improved the performance for queries which populate records with associations to other models which also have associations:
- Avoiding _.cloneDeep of the model's attributes definition for each of the records which should be populated, and instead using _.cloneDeep only the attributes which are associations.
- Avoiding "instrumenting" the model records for associations, if those records were retrieved only to populate the main records which were queried (so the associations wouldn't be used anyway).

These changes have decreased the processing time on my tests of querying a model which populates three associations to models which also have associations to around than 30% of what it was before.

### DIFF
--- a/lib/waterline/model/lib/model.js
+++ b/lib/waterline/model/lib/model.js
@@ -38,21 +38,27 @@ var Model = module.exports = function(attrs, options) {
   // Cast things that need to be cast
   this._cast(attrs);
 
-  // Build association getters and setters
-  this._defineAssociations();
+  // Check to see if associations are necessary for this record
+  if (! options.showJoins) {
 
-  // Attach attributes to the model instance
-  for(var key in attrs) {
-    this[key] = attrs[key];
-
-    if(this.associationsCache.hasOwnProperty(key)) {
-      this.associationsCache[key] = _.cloneDeep(attrs[key]);
-    }
+    // Build association getters and setters
+    this._defineAssociations();
   }
+    // Attach attributes to the model instance
+    for(var key in attrs) {
+      this[key] = attrs[key];
 
-  // Normalize associations
-  this._normalizeAssociations();
+      // Populate the associationsCache only if joins are needed and this property is an association
+      if(! options.showJoins && this.associationsCache.hasOwnProperty(key)) {
+        this.associationsCache[key] = _.cloneDeep(attrs[key]);
+      }
+    }
 
+  if (! options.showJoins) {
+
+    // Normalize associations
+    this._normalizeAssociations();
+  }
 
   /**
    * Log output
--- a/lib/waterline/model/lib/model.js
+++ b/lib/waterline/model/lib/model.js
@@ -39,8 +39,7 @@ var Model = module.exports = function(attrs, options) {
   this._cast(attrs);
 
   // Check to see if associations are necessary for this record
-  if (! options.showJoins) {
-
+  if (! options.ignoreAssociations) {
     // Build association getters and setters
     this._defineAssociations();
   }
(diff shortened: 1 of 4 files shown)


devinivy:
I would love this to go through.  It's a small, high-value PR.  Maybe someone should peek this closely to make sure it's safe, as this touches crucial pieces of code.  Were those `cloneDeep`s definitely unnecessary?  Would changes from this PR be confounding while implementing the ever-popular recursive-populate feature (#308)?

devinivy:
I looked through the changes.  My only concern is that it's often right but perhaps generally wrong to assume that you shouldn't be able to `add` or `remove` on populated records.  One way you're saving time is by,
> Avoiding "instrumenting" the model records for associations, if those records were retrieved only to populate the main records which were queried _(so the associations wouldn't be used anyway)_.

I imagine some use-case like this,
```
Directory.findOne({name: 'waterline2'}).populate('files')
.then(function(directory) {

    var file;
    for (var i = 0; i < directory.files.length; i++) {
        file = directory.files[i];
        file.collaborators.add(someId);
        file.save();
    }

});
```

nbiton:
@devinivy Thanks for the feedback.
You're right, this is a piece of functionality that gets broken by this PR.
Since you can't populate more than one level deep, I didn't think that this was possible before- as you can add and remove, but you don't get the currently associated item.

I think that if this functionality wasn't intended to work this way, this PR could make its way into a future release.

devinivy:
Yeah, I want the best of both worlds!  The performance increase would be really nice...

lib/waterline/model/lib/model.js
```diff
@@ -38,21 +38,26 @@ var Model = module.exports = function(attrs, options) {
...
-
-  // Attach attributes to the model instance
-  for(var key in attrs) {
-    this[key] = attrs[key];
-
-    if(this.associationsCache.hasOwnProperty(key)) {
-      this.associationsCache[key] = _.cloneDeep(attrs[key]);
-    }
+  // Check to see if associations are necessary for this record
+  if (! options.ignoreAssociations) {
```
tjwebb: There does this option come from? This needs docs.

dmarcelino:
This change does improve performance significantly, check results in [balderdashy/sails-mysq#161#issuecomment-88508058](https://github.com/balderdashy/sails-mysql/issues/161#issuecomment-88508058)

blah238:
This patch is great -- it improved the response time mentioned in https://github.com/balderdashy/sails-postgresql/issues/141#issue-64400447 from ~12 minutes to 16 seconds, and peak memory consumption from ~1.5GB to ~500MB. This is still too slow for production but it's a massive improvement.

blah238:
@dmarcelino @nbiton @tjwebb @devinivy What do we need to do to get this merged?

tjwebb:
Needs to be up-merged with master. @blah238 it looks like most of the heavy lifting is done, would you like to help us out by resolving some of the conflicts?

blah238:
@tjwebb The only conflict when I merged locally was the package.json version number change, so if you just merge https://github.com/nbiton/waterline/commit/316c3cff79103279ff2268132994f0752243551e then there should be no conflicts.

tjwebb:
Ok thanks man

blah238:
@tjwebb Here you go I just added another PR without the package.json change: https://github.com/balderdashy/waterline/pull/924

All credit to @nbiton of course!

devinivy:
Thanks much, @nbiton for the PR and @blah238 for facilitating!  Move over to #924 for show-time.

nbiton:
I'm happy to see this code helped others :smile: 

@devinivy as you pointed out earlier in this thread, it does break/remove some functionality that's available with the current version of waterline, even though all tests still pass successfully.

@tjwebb the option you were asking about is an ad-hoc option set only in https://github.com/nbiton/waterline/blob/316c3cff79103279ff2268132994f0752243551e/lib/waterline/query/finders/joins.js#L181-L202
This was made to be used only internally by waterline, and not as an option somebody could set externally- basically, I was piggybacking the *options* object to pass-in this flag for the model constructor.

I opted for a focused solution that didn't involve many changes to the code base, as I needed something stable enough to be used quickly in a production environment.
I'm sure there must be other ways to achieve even better results, without removing any existing functionality.

Issue #819, opened 2015-01-29
are there anyways to extend types in our own adapters?


shekhei:
There are cases where certain types in the adapter are not quite available in the types list, and they all end up getting changed to string, which, i reckon, not quite desirable.

If there isnt, I think there should at least be a way to extend in your own adapter?

dmarcelino:
Another related suggestion brought by @GaryLCoxJr on #665:

> I understand the waterline way which is to ensure all db adapters work.  Why don't we add support for the types a db uses, for example on Postgre uuid and decimal, and then for other adapters that dont use those types we just set a type it does use.  This way we can utilize the correct data types for a given database.
> 
> Example:
> Mongo doesn't have a uuid type, so that adapter might return a string
> ```javascript
> case "uuid":
>   return "string";
> ```
> 
> whereas on the postgre it would use that type.  This would keep things working the way waterline is designed and would allow us to use the correct data types.  I'm willing to add these types and update the existing adapters.
> 
> Update: Or we can just have a default catch all that returns a string if none of the case types matched.

devinivy:
This can generally be dealt with in-adapter for special cases like ids (for example, in sails-dynamodb the primary attribute defaults to a string type that is auto-filled by a UUID).  Otherwise, often is the case that a waterline type plus a validator will do the trick.  It might not be an ideal solution, but these things work.  Consider writing mixins for adapter-specific validations.

dmarcelino:
@shekhei came up with an interesting solution for this on PR #832, I'm not sure it's **the** solution but it's definitely an interesting approach. I would like to have a more *formal* way of defining both types and validations on adapters, I believe it's something that makes sense.

sailsbot:
Thanks for posting, @shekhei.  I'm a repo bot-- nice to meet you!

It has been 60 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

dmarcelino:
[...]

PR #832, opened 2015-02-09
Allowing types to be defined in each adapter

shekhei:
Continuing #827 here, original text:

> I do think this is quite important, there are cases where supported types might be a bit different with each adapter.
> 
> With this change, this will allow each adapter to extend and create their own supported types.
> 
> Although this might mean that it becomes more difficult to have models that supports all sorts of different adapters, the chances of the models changing adapter is rather low, this change can make waterline more extensive :)

### DIFF
--- a/lib/waterline/adapter/index.js
+++ b/lib/waterline/adapter/index.js
@@ -33,5 +33,6 @@ _.extend(
   require('./aggregateQueries'),
   require('./setupTeardown'),
   require('./sync'),
-  require('./stream')
+  require('./stream'),
+  require('./types')
 );
--- a/lib/waterline/adapter/types.js
+++ b/lib/waterline/adapter/types.js
@@ -1,0 +1,6 @@
+module.exports = {
+  supportsType: function(type) {
+    var adapter = this.connections[connName]._adapter;
+    return adapter.supportsType && adapter.supportsType(type);
+  }
+};
--- a/lib/waterline/adapter/types.js
+++ b/lib/waterline/adapter/types.js
@@ -1,6 +1,7 @@
+var _ = require('lodash');
 module.exports = {
   supportsType: function(type) {
     var adapter = this.connections[connName]._adapter;
-    return adapter.supportsType && adapter.supportsType(type);
+    return _.isFunction(adapter.supportsType) && adapter.supportsType(type);
   }
 };
--- a/lib/waterline/core/index.js
+++ b/lib/waterline/core/index.js
@@ -86,7 +86,7 @@ Core._initialize = function(options) {
   var reservedAttributes = adapterInfo.reservedAttributes || {};
 
   // Initialize internal objects from attributes
-  this._schema.initialize(this._attributes, this.hasSchema, reservedAttributes);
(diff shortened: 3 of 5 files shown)


dmarcelino:
For reference, this fixes #819 and relates to #665.

dmarcelino:
@shekhei, can you post examples on how this would work? I guess this would also require changes on the adapters to actually support the non compliant (by waterline standards) types, right?

shekhei:
Basically, you can see it here
https://github.com/eskygo/sql-layer-adapter-sails/blob/master/lib/adapter.js#L62
If it is not defined, the default list is used, if this function is defined, it will result in a union of the default set and the returned set, combined with
https://github.com/eskygo/sql-layer-adapter-sails/blob/master/lib/adapter.js#L62
which allows it to be translated to actual types.

There should be no changes required for the current adapters.

dmarcelino:
Well, for an adapter to extend the waterline supported types it would need to implement `.supportsType(type)`, that was the bit of information I was after.

Any sort of validation would also be the adapter's responsibility and this change does not prescribe a way to do it, which is also something we should consider...

I do agree we need to provide a way for adapters to extend the supported types, I'm not 100% sure this is the best way. Ideally I would like to see something that would also offer validation support consistent with the current waterline validations (not sure if possible).

@devinivy, @tjwebb, @particlebanana, what do you think of this?

shekhei:
@dmarcelino absolutely agree that this method isn't exactly the most robust method. perhaps another way would be to allow types to be "defined". the current type mechanism in waterline seem to be only used for "type matching" and "DDL translation". 

A better way might be to do something similar to, perhaps mongoose where type definition are objects? But of course that will require quite a bit of change.

dmarcelino:
@shekhei, I've just stumbled upon PR #72 which reads custom types from collection and this got me thinking... what if we extend this so waterline core reads custom types from the adapter? The idea would be, in adapter.js we would optionally add a `types` object like:
``` javascript
  //custom types
  types:{
     geoJSON: isGeoJSON_function
  },
```

If present waterline core would read it and apply those extra custom types. That way we would satisfy both the schema and validation sides of this problem.

What do you think?

shekhei:
That sounds good! 

This "type object" would probably end up being something complex though, any thoughts on that? I've seen how mongoose does it, very robust, but it gets rather complex quickly.

https://github.com/Automattic/mongoose/blob/master/lib/types/array.js

dmarcelino:
I guess that would be a "problem" for each adapter to solve but we can follow anchor's [rules.js](https://github.com/balderdashy/anchor/blob/master/lib/match/rules.js) where the functions are actually defined in a different file.

shekhei:
Any examples on how you reckon it would end up looking like? Perhaps for a standard type that's already supported?

dmarcelino:
Anchor is what waterline uses under the covers, take a look at: https://github.com/balderdashy/anchor/blob/master/lib/match/rules.js#L55

dmarcelino:
@shekhei, take a look at #977, I think it's the best way to address this. What do you think?

dmarcelino:
Note, PR #977 was replaced by #981.

dmarcelino:
Closing this in favor of #981.

PR #837, opened 2015-02-12
[WiP] Lodash 3 refactors

megawac:
Please review before further work continues...

~~Was working on this over the past 40 minutes or so, let me know if I should continue.~~

Note to self
  - refactor query/finders/basic `find` and `findOne` in a separate PR

### DIFF
--- a/lib/waterline/utils/normalize.js
+++ b/lib/waterline/utils/normalize.js
@@ -17,7 +17,7 @@ var normalize = module.exports = {
     if (!context.autoPK) {
       // Check which attribute is used as primary key
       for(var key in context.attributes) {
-        if(!util.object.hasOwnProperty(context.attributes[key], 'primaryKey')) continue;
+        if(!_.has(context.attributes[key], 'primaryKey')) continue;
 
         // Check if custom primaryKey value is falsy
         if(!context.attributes[key].primaryKey) continue;
@@ -30,7 +30,7 @@ var normalize = module.exports = {
 
     // Check if options is an integer or string and normalize criteria
     // to object, using the specified primary key field.
-    if(_.isNumber(options) || _.isString(options) || Array.isArray(options)) {
+    if(_.isNumber(options) || _.isString(options) || _.isArray(options)) {
       // Temporary store the given criteria
       var pkCriteria = _.clone(options);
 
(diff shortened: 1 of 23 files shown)


Globegitter:
:+1: I would love to see that. One thing I would suggest though is requiring `lodash` modules directly, at least where it makes sense.
So for example you can now do 
```js
var isObject = require('lodash/Lang/isObject');
var reduce = require('lodash/Collection/reduce');
```
which can improve the performance.

tjwebb:
+1

lib/waterline/utils/normalize.js
```diff
@@ -192,11 +192,8 @@ var normalize = module.exports = {
     // empty object without running the query because nothing will match anyway. Let's return
     // false from here so the query knows to exit out.
     if(criteria.where) {
-      var falsy = false;
-      Object.keys(criteria.where).forEach(function(key) {
-        if(Array.isArray(criteria.where[key]) && criteria.where[key].length === 0) {
-          falsy = true;
-        }
+      var falsy = _.any(criteria.where, function(prop) {
+        return _.isArray(prop) && prop.length === 0;
```
megawac: This can probably just be `_.any(criteria.where, _.isEmpty)`

megawac:
> One thing I would suggest though is requiring lodash modules directly, at least where it makes sense. [...] which can improve performance

In this case we're using so much of lodash that I doubt that will make much sense. It would more likely degrade performance requiring each module piece by piece whereas this is using the prebuilt `index.js`

Can people make sure I'm not breaking anything before this PR gets too out of hand.

Globegitter:
@megawac Not so sure about that: https://github.com/babel/babel switched from requiring `_` all over the place to modules directly and `requiring` time dropped quite significantly. From about 2s to below 1s. There may have been some more performance improvements, but lodash was certainly one of the biggest factors.

And yeah will test in my local project asap.

listepo:
+1

megawac:
> :+1: I would love to see that. One thing I would suggest though is requiring lodash modules directly

@Globegitter made a tool a little while ago pretty much based on your comment https://github.com/megawac/lodash-modularize

particlebanana:
There is just too much to do at once. As pieces get refactored we can move to more lodash features but doing it all in one pass doesn't seem worth it and would be too much to try and verify the pr against.

PR #843, opened 2015-02-18
Deprecate findOrCreateEach

Globegitter:
According to http://stackoverflow.com/questions/27573734/how-do-i-use-findorcreateeach-in-waterline-sailsjs findOrCreateEach is deprecated.

Also I am not finding much about `findOneLike` or `findLike`. Are these deprecated as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ Each of the following basic methods are available by default on a Collection ins
 In addition you also have the following helper methods:
 
   - createEach
-  - findOrCreateEach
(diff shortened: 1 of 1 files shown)


tjwebb:
Good question. Personally I'd like to deprecate (and remove) all dynamic finders, since it's just senseless proliferation of API surface area.

particlebanana:
I would love to remove the dynamic finders. Whenever we feel like an 0.11 release is ready we can take a pass and remove all that stuff.

PR #845, opened 2015-02-20
fixed findOrCreate and added more tests.

Globegitter:
When I had `Topic.findOrCreate({ researchId:  meta.id, text: meta.text }).exec()` as well as `Topic.findOrCreate([{ researchId:  meta.id, text: meta.text }, { researchId:  'bladi', text:  'kaaa' }]).exec()` in my app with https://github.com/balderdashy/sails-disk it just did not work properly.

Fixed that, added some more tests and changed `forEachSeries` to `eachSeries` since that seems to be the preferred name (now?) by the library https://github.com/caolan/async#eachseriesarr-iterator-callback

### DIFF
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -43,7 +43,7 @@ module.exports = {
 
     if(!hasOwnProperty(adapter, 'create')) return cb(new Error(err));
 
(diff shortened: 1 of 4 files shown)


sgress454:
This looks delightful, @particlebanana whaddayasay?

particlebanana:
:+1: thanks @Globegitter

Globegitter:
Thanks for the quick merge :)

Issue #533, opened 2014-07-18
crashes when trying to update a model with a new association-record-to-create


mkeremguc:
title is self explanatory i believe.. here is the error i get:
```
/usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:532
    this.create(record).exec(function(err, val) {
         ^
TypeError: Object #<Object> has no method 'create'
    at createRecord (/usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:532:10)
    at createRunner (/usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:521:20)
    at /usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:125:13
    at Array.forEach (native)
    at _each (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:46:24)
    at Object.async.each (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:124:9)
    at mapAssociations (/usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:524:11)
    at /usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:125:13
    at Array.forEach (native)
    at _each (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:46:24)
```

paquettea:
Thanks for sharing you investigation in details. I encountered the same issue, and the fix you are suggesting seems to be properly working ... I hope we will have a followup on this soon.

cratervale:
Bump

mkeremguc:
bump

mkeremguc:
can you PR already ?

paquettea:
https://github.com/balderdashy/waterline/pull/850

tjwebb:
> can you PR already ?

If you're too lazy to submit your own patch, you're not allowed to complain.

mkeremguc:
@tjwebb paquettea already committed my suggestion on his own repo so what is wrong here to ask him to PR ? moreover, i'm not complaining

devinivy:
See #934 .

Issue #644, opened 2014-09-17
nestedOperations update Problem. 


geilt:
Using the following attributes and passing in an array of mongoids. Trying like hell to store mongo ids in an array.

investors: {
   model: 'User',
   type: 'array'
},

/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:532
    this.create(record).exec(function(err, val) {
         ^
TypeError: Object #<Object> has no method 'create'
    at createRecord (/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:532:10)
    at createRunner (/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:521:20)
    at /node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:125:13
    at Array.forEach (native)
    at _each (/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:46:24)
    at Object.async.each (/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:124:9)
    at mapAssociations (/node_modules/sails/node_modules/waterline/lib/waterline/utils/nestedOperations/update.js:524:11)
    at /node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:125:13
    at Array.forEach (native)
    at _each (/node_modules/sails/node_modules/waterline/node_modules/async/lib/async.js:46:24)

devinivy:
Does this issue remain?

mkeremguc:
yes. https://github.com/balderdashy/waterline/issues/533

dmarcelino:
Hi @mkeremguc, there have been two PRs (#889 and #898) recently released that address nested updates. Can you try the latest version of waterline and check if the issue remains? Thanks

mkeremguc:
Hi @dmarcelino, unfortunately the issue still remains on latest waterline(0.10.20). can you please check my observation and possible fix within #533 ?

mkeremguc:
obviously this: https://github.com/balderdashy/waterline/issues/533#issuecomment-68079748

dmarcelino:
@mkeremguc, thanks for checking. The PRs I mentioned fixed several issues with nested updates and I wanted to be sure that it didn't fix this issue.

I've left a comment on PR #850 which is supposedly the fix for this issue.

mkeremguc:
PR #930

dmarcelino:
[...]

PR #850, opened 2015-02-22
patch suggested by mkeremguc https://github.com/balderdashy/waterline/issues/533

paquettea:
Here's the PR but I haven't done tests.. I don't know what the waterline team thinks about this fix since we've haven't had much feedback ...


dmarcelino:
Hi @paquettea, would you mind adding a test to this PR? Something that fails without the fix and passes when the fix is applied. Thanks

cc: @mkeremguc (regarding #644)

mkeremguc:
@dmarcelino, #930

PR #853, opened 2015-02-23
#846: add migration strategy `create`

dmarcelino:
This PR adds support for `create` migration strategy. It differs from `alter` by only running secure operations as `.define()` and `.addAttribute()` (if adapter implements it). More details in #846.

Tests added in balderdashy/waterline-adapter-tests#34

### DIFF
--- a/lib/waterline/adapter/sync/index.js
+++ b/lib/waterline/adapter/sync/index.js
@@ -2,5 +2,6 @@
 module.exports = {
   migrateDrop: require('./strategies/drop.js'),
   migrateAlter: require('./strategies/alter.js'),
+  migrateCreate: require('./strategies/create.js'),
   migrateSafe: require('./strategies/safe.js')
 };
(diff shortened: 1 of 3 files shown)


dmarcelino:
@particlebanana, any concerns regarding this? I'm open to discuss and change the PR. Let me know your thoughts. Cheers

particlebanana:
My only concern is that when dealing with anything on a production database it's clear that the developer knows exactly what is happening. When you have some process adding columns to the database there is a sense that something magical is happening. With migrations you have a deliberate action being taken to modify the database and a way to easily rollback changes. It also allows you to do things like set default values for existing records, etc. 

This is why we force `migrate: safe` in Sails when `NODE_ENV=production` so that we can be sure we don't modify your production schema.

I'm cool with adding this in to Waterline though. I think we should keep Sails the way it is at least for the immediate term. Thanks!

dmarcelino:
Thanks @particlebanana. I see what you mean and that's why I consider this migration *safish*, not really safe.

One conundrum I faced while writing this was what to do with default values when adding columns. Is this something that should be handled by waterline core or by the adapter's `addAttribute` method? 

There still isn't much documentation about `migratable` which makes it hard to contribute to... Hopefully that will clear up by 0.11? :)

Thanks again.

Issue #858, opened 2015-02-25
Remove/save behaving very incorrectly for many-to-many relationship.


devinivy:
In a rush... I will fill this out more soon.

But `remove` with a `save` is basically removing _all_ associated records instead of just those specified.  I'm using sails-mysql, but I've found that the issue lives in waterline itself.  `save` triggers a removal of many-to-many records correctly, but the nested operations for `save`'s implicit `update` is completely destroying all of a person's friends (per example below).  There seems to be a sync of join table data going awry.  I was bummed to see that approach to keeping nested elements up-to-date– it cannot work for non-transactional queries, which waterline does not support.  And it's prone to error when racing against a possibly simultaneous `remove` operation.

```
person.friends.remove(1);
person.save(function() {
    // now person has no friends
});
```

If you use many-to-many associations, this affects you!  Would love to get a PR up today.  Workin' on it, but would love some help.

devinivy:
Update...
Seems this happens when `person` is populated with a subset of `friends`.  Only that subset will be restored during the nested operations update.  Definitely a huge bug!  @particlebanana what's the purpose of adding/removing relations during an update?  It not documented and _seems_ unnecessary to sync, but I'm curious to get some input.  I do agree that updating the values of nested objects makes sense, but I don't think that syncing the relationships themselves is a good idea, especially via syncing.  Code example of the issue:

```
Person.findOne(666).populate('friends', {id: 42})
.then(function(person) {
    person.friends.remove(42);
    person.save() {
        // now I have no friends, since nested operations assumed that this person had only one friend to begin with (friend 42), which is false.
    }
})
```

devinivy:
Relevant stack trace of the save/update causing this,
```
    at module.exports [as update] (XXX/node_modules/waterline/lib/waterline/query/dql/update.js:29:13)
    at new module.exports (XXX/node_modules/waterline/lib/waterline/model/lib/associationMethods/update.js:61:14)
    at Array.async.auto.updateRecord [as 1] (XXX/node_modules/waterline/lib/waterline/model/lib/defaultMethods/save.js:73:41)
    at XXX/node_modules/waterline/node_modules/async/lib/async.js:487:38
    at Array.forEach (native)
    at _each (XXX/node_modules/waterline/node_modules/async/lib/async.js:46:24)
    at Object.async.auto (XXX/node_modules/waterline/node_modules/async/lib/async.js:458:9)
[...]

PR #861, opened 2015-02-26
Update related models but do not sync relationships during a nested update.

devinivy:
Fixes bug #858 .  This PR touches some crucial pieces of code, so it would be nice to have help testing this out to make sure it works properly.  That said, mocha is happy with the results.  A big upside is that the number of queries is effectively halved on updates to associated records.

The "downside" is that manipulating `add`s and `remove`s manually by removing and adding items from a model's association array does not work when running `save`.  It is necessary to use `add` and `remove` now.  Bear in mind that this piece of functionality that is being dropped is not documented and was not really tested, so I don't personally consider it part of the waterline API.  That means I expect this change can go into a 0.10.x release with some sort of caveat posted.

Nested models should still have updates applied to them.

### DIFF
--- a/lib/waterline/utils/nestedOperations/update.js
+++ b/lib/waterline/utils/nestedOperations/update.js
@@ -32,10 +32,8 @@ module.exports = function(parents, values, associations, cb) {
   var operations = buildOperations.call(self, associations, values);
 
   // Now that our operations are built, lets go through and run any updates.
-  // Then for each parent, find all the current associations and remove them then add
-  // all the new associations in using .add()
-  sync.call(self, operations, cb);
-
+  updateRunner.call(self, operations, cb);
+  
 };
 
 
@@ -61,10 +59,9 @@ function buildOperations(associations, values) {
     // keys on the parent model.
     if(optValues === null) return;
 
-    // Pull out any association values that have primary keys, these will need to be updated. All
-    // values can be added for each parent however.
+    // Currently just performs updates, but we can create other runners and resort to async
+    // if other operations should also happen.
     operations[association] = {
-      add: [],
       update: []
     };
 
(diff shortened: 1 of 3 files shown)


devinivy:
Looking at #829, wondering if this is an issue...
```
Directory.findOne({name: 'waterline2'}).populate('files')
.then(function(directory) {

    var file;
    for (var i = 0; i < directory.files.length; i++) {
        file = directory.files[i];
        file.collaborators.add(someId);
        file.save();
    }

});
```

Perhaps the nested updates should use `Model.save` rather than `Collection.update`.  Thoughts?

devinivy:
Has anyone peeked at this yet?  It's fairly crucial issue for us, so I'm trying to evaluate whether or not we should migrate over to my fork of waterline.

particlebanana:
@devinivy this is failing the waterline adapter tests for association updates. ex https://github.com/balderdashy/waterline-adapter-tests/blob/master/interfaces/associations/hasMany/update.nested.js#L37-L62

Is there anyway we can focus this down to just making https://github.com/balderdashy/waterline/blob/master/lib/waterline/model/lib/defaultMethods/save.js#L68-L76 and https://github.com/balderdashy/waterline/blob/master/lib/waterline/model/lib/associationMethods/update.js#L22-L62 not handle nested updates that are not in the `.add` and `.remove`? Just in the save do a `toObject` and then remove any objects or arrays before doing the update? That would allow `.add` and `.remove` to continue working but fix any issues with changing populated associations.

If we are going to take a hack saw to `update` we might as well pull out all the nested stuff. That would take more time though and we need to get something out.

Let me know if you already looked at that though.

particlebanana:
cc @Latros

Dallas62:
@devinivy
```
Directory.findOne({name: 'waterline2'}).populate('files')
.then(function(directory) {

    var file;
    for (var i = 0; i < directory.files.length; i++) {
        file = directory.files[i];
        file.collaborators.add(someId);
        file.save();
    }

});
```
I think this case is not optimised, by saving each times, but can happen if there is an action's of a controller call many times at the same time (over HTTP).
First case of https://github.com/balderdashy/waterline/issues/877

devinivy:
@particlebanana I imagine something like that could work, but it's actually pretty hard (for me) to figure out precisely the right way to do it.  I'd like the special case of `updates` explicitly resetting associations to look and feel like a special case in the code.  There really cannot be any other time that every association is dropped, and that's clearly an issue right now– a somewhat confounding one, even!  Just want to keep that data safe!

I'll continue down the route you suggested, but any help is still appreciated.

devinivy:
Closing this out without merge, as it simply isn't the solution while supporting adapter tests properly.  Perhaps some day we can reopen.

Issue #835, opened 2015-02-11
time data type validation


antoniosap:
Hi all,

inserting a record in my model 'Crono':
...
  	timeStart: 	  { type: 'time' }, 
  	timeEnd: 	  { type: 'time' },  
...

....with js code ....
timeStart = new Date().getTime();
timeEnd = new Date().getTime();

give me validation error:

Invalid attributes sent to Crono:
 • timeStart
   • [object Object]
 • timeEnd
   • [object Object]

gotmikhail:
https://github.com/balderdashy/waterline/issues/826

I reported this as a compatibility issue w/ `moment.js`. Looks like it might actually be a compatibility issue w/ the js `Date` object.

gsdatta:
Getting the same issue, with postgres adapter if that helps.

antoniosap:
also redis adapter

gotmikhail:
Mine was w/ the mysql adapter

gsdatta:
Looks like it's not even an issue with the Date object. I tried using moment to parse the input value (rather than Date.parse()), and returned the Date representation of the object if the moment object was valid. Even with what I'm sure is a valid object, I'm receiving the same validation error. 

```
Error (E_VALIDATION) :: 2 attributes are invalid
    at WLValidationError.WLError (/Volumes/Storage/Development/myproject/node_modules/waterline/lib/waterline/error/WLError.js:33:18)
    at new WLValidationError (/Volumes/Storage/Development/myproject/node_modules/waterline/lib/waterline/error/WLValidationError.js:20:28)
    at /Volumes/Storage/Development/myproject/node_modules/waterline/lib/waterline/query/validate.js:45:43
    at allValidationsChecked (/Volumes/Storage/Development/myproject/node_modules/waterline/lib/waterline/core/validations.js:203:5)
    at done (/Volumes/Storage/Development/myproject/node_modules/waterline/node_modules/async/lib/async.js:135:19)
    at /Volumes/Storage/Development/myproject/node_modules/waterline/node_modules/async/lib/async.js:32:16
    at validate (/Volumes/Storage/Development/myproject/node_modules/waterline/lib/waterline/core/validations.js:135:49)
    at /Volumes/Storage/Development/myproject/node_modules/waterline/node_modules/async/lib/async.js:125:13
    at Array.forEach (native)
    at _each (/Volumes/Storage/Development/myproject/node_modules/waterline/node_modules/async/lib/async.js:46:24)
[...]

PR #863, opened 2015-02-27
Fixed time validation issue (on the waterline side).

gsdatta:
Added a typecast for time that parses a time (e.g. 1:00:00 PM), etc. Returns this in a standardized time format understood by databases (MySQL, Postgres, and others), which is "hh:mm:ss a", where a is AM/PM. However, the MAIN issue is in the anchor library used by waterline, and I am submitting a pull request there as well.

Moment.js is needed for this fix (its parsing capabilities are far higher than Date's).

### DIFF
--- a/lib/waterline/core/typecast.js
+++ b/lib/waterline/core/typecast.js
@@ -216,7 +216,7 @@ Cast.prototype.time = function time(value){
   }
 
   if(_value.isValid()){
-    return _value.toISOString();
+    return _value.format("hh:mm:ss A");
   }
   else
     return value;
--- a/lib/waterline/core/typecast.js
+++ b/lib/waterline/core/typecast.js
@@ -6,6 +6,7 @@ var types = require('../utils/types'),
     utils = require('../utils/helpers'),
     hasOwnProperty = utils.object.hasOwnProperty,
     _ = require('lodash');
+    var moment = require('moment');
 
 /**
  * Cast Types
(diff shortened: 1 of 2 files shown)


gsdatta:
Bump! This is something pretty basic to the functionality of an ORM, and needs to be addressed whether by this fix or something else!

lib/waterline/core/typecast.js
```diff
@@ -87,11 +88,14 @@ Cast.prototype.run = function(values) {
         break;
 
       case 'date':
-      case 'time':
```
dmarcelino: Should `time` have a different logic than `datetime` and `date`? That looks like an API change.

dmarcelino:
For reference, this PR addresses #835.

particlebanana:
Hmm I don't know what I think about this. Time is a tough one and I've never used it to store data that way so I'm not a good person to answer this. Do you usually include a date in another field in the db?

I guess my original thinking with `datetime`, `date` and `time` is that it's easy to use a `datetime` field in the adapter and you can always grab the time out of it or just the date if you prefer. If you hate dealing with timezone calculations like me use milliseconds or seconds since epoch. 

I don't think the WL query language supports querying times using `hh:mm:ss` either unless it's been added.

I agree it's confusing to have all three and have them not perform the way you would expect. Just seems hard to port across adapters, I think stuff like mongo only maps full JS date objects right?

cc: @tjwebb

dmarcelino:
@particlebanana, @tjwebb, there is a bit more discussion about this going on balderdashy/anchor#69.

particlebanana:
This has conflicts and can't be merged. I don't feel strongly one way or the other but there doesn't seem to be enough input from the community on how `time` is used. 

I appreciate your work here @gsdatta and if you still feel strongly that we need more fine grained casting of `time` attributes following the [Contribution Guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) and submitting a proposal would be more than welcome.

PR #866, opened 2015-02-27
Add stack and message properties to waterline errors

ansel1:
Fix for #865.  Makes waterline errors behave more like Error.  In particular, mocha tests which throw waterline errors will now report the error stack instead of just reporting nothing.


particlebanana:
Awesome thanks!

PR #889, opened 2015-03-17
Disable parent cache on update nested operations

### LINKED ISSUE #877: [CRITICAL] Data corruption
Hi!

When i'm using Sails with sails-mysql, I have some data corruption on "High" frequency update...

The first one is that we cannot change associations at the same time.
ex:
```
Model.findOne(1).exec(function(err, model){
model.pets.add( 1 );
model.save();
});
Model.findOne(1).exec(function(err, model){
model.pets.add( 2 );
model.save();
});
Model.findOne(1).exec(function(err, model){
model.pets.add( 3 );
model.save();
});

// Expected on Joining Table:
// 1 - 1
// 1 - 2
// 1 - 3

// Result on Joining Table:
// 1 - 1
// Or
// 1 - 2
// Or
// 1 - 3
// Or
// 1 - 1
// 1 - 2
// Or
// 1 - 1
// 1 - 3
// Or
// 1 - 2
// 1 - 3
// Or ...
```
This could be done by HTTP request, so it's critical

Same with update:
```
Model.update(1, { pets: [1] }).exec();
Model.update(3, { pets: [1] }).exec();
Model.update(4, { pets: [1] }).exec();
// Expected on Joining Table:
// 1 - 1
// 3 - 1
// 4 - 1

// Result on Joining Table:
// 1 - 1
// 1 - 1
// 1 - 1
// Or
// 3 - 1
// 3 - 1
// 3 - 1
// Or
// 4 - 1
// 4 - 1
// 4 - 1
// Or
// 1 - 1
// 3 - 1
// 4 - 1
// Or
// 1 - 1
// 3 - 1
// Missing 4 - 1
// Or
// Missing 1 - 1
// 3 - 1 
// 4 - 1
// ...
// Math.random();
```
This could be done by HTTP request, so it's critical too

INFOS:
sails:           0.11.0
waterline:    0.10.18
sails-mysql  0.10.11

EDIT: On Many-To-Many

@mikermcneil @sgress454

devinivy:
Should resolve #877

### DIFF
--- a/lib/waterline/utils/nestedOperations/update.js
+++ b/lib/waterline/utils/nestedOperations/update.js
@@ -22,19 +22,16 @@ module.exports = function(parents, values, associations, cb) {
 
   var self = this;
 
-  // Cache parents
-  this.parents = parents;
-
   // Combine model and collection associations
   associations = associations.collections.concat(associations.models);
 
   // Build up .add and .update operations for each association
-  var operations = buildOperations.call(self, associations, values);
+  var operations = buildOperations.call(self, parents, associations, values);
 
   // Now that our operations are built, lets go through and run any updates.
   // Then for each parent, find all the current associations and remove them then add
   // all the new associations in using .add()
-  sync.call(self, operations, cb);
+  sync.call(self, parents, operations, cb);
 
 };
 
@@ -47,7 +44,7 @@ module.exports = function(parents, values, associations, cb) {
  * @return {Object}
  */
 
-function buildOperations(associations, values) {
+function buildOperations(parents, associations, values) {
 
   var self = this;
   var operations = {};
@@ -70,7 +67,7 @@ function buildOperations(associations, values) {
 
     // Normalize optValues to an array
     if(!Array.isArray(optValues)) optValues = [optValues];
-    queueOperations.call(self, association, operations[association], optValues);
+    queueOperations.call(self, parents, association, operations[association], optValues);
   });
 
   return operations;
@@ -87,7 +84,7 @@ function buildOperations(associations, values) {
  * @param {Array} values
  */
 
-function queueOperations(association, operation, values) {
+function queueOperations(parents, association, operation, values) {
 
   var self = this;
   var attribute = self.waterline.schema[self.identity].attributes[association];
@@ -105,7 +102,7 @@ function queueOperations(association, operation, values) {
   if(collection.junctionTable) {
 
     // For each parent, queue up any .add() operations
-    self.parents.forEach(function(parent) {
+    parents.forEach(function(parent) {
       values.forEach(function(val) {
         if(!hop(parent, association)) return;
         if(typeof parent[association].add !== 'function') return;
@@ -141,7 +138,7 @@ function queueOperations(association, operation, values) {
     }
 
     // Set the new foreign key value for each parent
-    self.parents.forEach(function(parent) {
+    parents.forEach(function(parent) {
       parent[association] = val[modelPk];
     });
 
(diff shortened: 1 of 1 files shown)


devinivy:
Possibly fixes #911

devinivy:
@dmarcelino @tjwebb @particlebanana we've received a couple issues related to this, and it's quite important for data integrity.  Anyone mind taking a peek?

tjwebb:
For reference can you describe how disabling the parent ids fixes the high-frequency update issue?

devinivy:
During a `Collection#update` (which occurs also during `Model#save`), "nested operations" are run to sync associated records.  That is, if I execute,
`Project.update({status: "unmerged"}, {collaborators: [1,3,5]}).exec(function(){});`,  all projects with the "unmerged" status are assigned exactly collaborators 1, 3, and, 5.  That includes unlinking any other collaborators from all projects with that status, then making sure links are made for the appropriate collaborators.  In this case, all projects with status "unmerged" are the `parents` referred to in this PR.

The nested operations store `parents` on the function object then perform some asynchronous operations.  Although the parents are stored on the function object, the nested operations are not called as a `new` instance (https://github.com/balderdashy/waterline/blob/be250e4aa1c0134e1c3a9c16650a3cabf0bacd23/lib/waterline/query/dql/update.js#L262).  That means that the parents can be shared from several separate `Collection#update` operations.  To mitigate this, I passed the parents around in function arguments just like everything else gets passed-around during the nested operations.

dmarcelino:
Hey @devinivy, is this something hard to test? I would feel more comfortable with a failing test and then see that the fix passes the test. My reasons are threefold:
- Confirm the issue is fixed;
- Prevent a regression in the future;
- I'm not yet too comfortable with the core code :).

devinivy:
The test for this would probably live in the adapter tests.  I can write one pretty easily, I think, but the test would have to be against something like sails-memory.

dmarcelino:
I think that's perfectly fine! Besides, waterline core already runs the sails-memory tests (edge version) as part of its test suite. Check the waterline core test/adapter folder ;)

devinivy:
@dmarcelino see PR https://github.com/balderdashy/waterline-adapter-tests/pull/42

dmarcelino:
Thanks for writing the test @devinivy. I've checked out your waterline-adapter-tests branch, ran it and could see your test would fail. Then I've checked-out this waterline branch, linked it and could see it fixed your test. From what I'm concerned both this PR and balderdashy/waterline-adapter-tests#42 are good to be merged! :white_check_mark:

I'll let @tjwebb take a look.

RWOverdijk:
:+1: In solving the original issue
:-1: For the method chosen to fix the issue. I'd scope the updates per parent cluster to avoid collisions (now it's a broader scope when debugging, figuring out what layer you're in).

devinivy:
What you're saying is reasonable, but does not really follow the patterns used there, which is what I was catering to.  The `values` / `associations` are just as important as the `parents`, and they get passed around functionally.  Restructuring this module would make for a great PR.  In the meantime, this is an easy-to-understand fix for a critical issue, following the conventions used in the module.

RWOverdijk:
@devinivy That's why I did grant you my :+1: :) I just felt like this had to be said.

devinivy:
:ok_hand:

particlebanana:
merged and published `0.10.20`

dmarcelino:
Nice! :smiley: 

@particlebanana please take a look at balderdashy/waterline-sequel#34, it needs your magic touch to enable waterline-sequel in https://circleci.com so we can use it for integration tests!

Issue #887, opened 2015-03-15
migrate: 'alter' when using Waterline standalone


davidhq:
I got bitten by this gotcha and lost (development) data...

I and some others on gitter chat think default should be `migrate: 'safe'`

https://github.com/balderdashy/waterline/blob/63effdb480ae9bd61cd78fd534e1e513294bd98d/lib/waterline/collection/defaults.js

@dmarcelino edit: this issue has been split into 2 parts (check the below [comment](https://github.com/balderdashy/waterline/issues/887#issuecomment-84045351)):
 - [x] Prevent users from using alter/drop in production by throwing an error. That way they can fix their config: #893
 - [ ] Change default migration to `'safe'` (or `'create'`)

robertbak:
Applies to production to.
See this TODO`s: 
https://github.com/balderdashy/waterline/blob/25ccb99e92ea018ca832bd4497af903c4212b34e/lib/waterline/adapter/sync/strategies/alter.js#L22
https://github.com/balderdashy/waterline/blob/25ccb99e92ea018ca832bd4497af903c4212b34e/lib/waterline/adapter/sync/strategies/drop.js#L20

davidhq:
Something else I'd like to add... since I'm using this separately from Sails, I will use the exact same code / setup on production... 

So I think this really should not be the default...

dmarcelino:
I believe a good candidate for a default migration strategy in Waterline standalone is `migrate: 'create'` (#846). It's on master branch and it hasn't been released yet. `'create'` has the advantage of creating tables when one first launches waterline and it never deletes any data. It will also create any missing tables/columns in subsequent initialisations. It can be seen as *safeish* alter migration.

devinivy:
I would vote for `'safe'`

particlebanana:
I'm cool with either `safe` or `create`. The tricky part is when you start with:

```javascript
username: {
    type: 'string'
}
```

and then change it later to:

```javascript
username: {
    type: 'string',
    unique: true
}
```
With `alter` it would pull the data out and make sure these indexes get built. To be honest for sql databases we switched to migrations long ago on our projects so mine is set to `migrate: 'safe'` from the start but I'll let you guys choose whatever works best for your use cases and we can set that as the default. Losing data is bad and if what we have now isn't working for people it's an easy change.

dmarcelino:
[...]

PR #893, opened 2015-03-20
#887: Refuse to run alter/drop migration strategies in production.

dmarcelino:
Addresses issue #887.

### DIFF
--- a/lib/waterline/adapter/sync/strategies/alter.js
+++ b/lib/waterline/adapter/sync/strategies/alter.js
@@ -19,7 +19,7 @@ module.exports = function(cb) {
 
   // Refuse to run this migration strategy in production.
   if (process.env.NODE_ENV === 'production'){
-    return cb();
+    return cb(new Error('`migrate: "alter"` strategy is not supported in production, please change to `migrate: "safe"`.'));
   }
 
   // Find any junctionTables that reference this collection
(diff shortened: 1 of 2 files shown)


devinivy:
Sweet– can we also change the default strategy on production to `"safe"` and submit a PR to waterline-adapter-tests that would go up with this PR in tandem?

dmarcelino:
Hey @devinivy, I've tried changing the default to `'safe'` in my machine but it breaks the waterline-adapter-tests. I've mentioned that here: https://github.com/balderdashy/waterline/issues/887#issuecomment-84332762

devinivy:
Yup– that's why I'm suggesting we also submit a PR to waterline-adapter-tests that will allow the change to occur without breaking any tests.  Or are you saying that we should not change the adapter tests to cater to this PR?

dmarcelino:
Ah, I see what you mean now, sorry. That should be the way to do it but honestly I haven't yet looked at what needs to be changed in `waterline-adapter-tests`. I suppose several fixtures need to be changed to `migrate: 'alter'`. For I now I would prefer to keep this PR and the `migrate: 'safe'` default as different PRs as they can be merged independently.

devinivy:
Good enough for me!  It would be great if we could work to have the adapter tests, docs, and default migration strategy all updated soon after this PR goes through.

dmarcelino:
Agreed! And stage 2 of that would be to discuss the future of migrations (both automated (dev) and manual(manual)). There are some interesting API ideas but they never developed on them, check this:
https://github.com/balderdashy/sails-docs/blob/master/contributing/adapter-specification.md#migratable-interface

dmarcelino:
Thanks @tjwebb!

Issue #862, opened 2015-02-26
Please add more collaborators


aaronshaf:
There seem to be quite a few outstanding pull requests, etc.

dmarcelino:
The outstanding issues / pull requests is also something that also worries me... Currently, there are 'a lot' of people willing to contribute but it seems multiple PR's are not getting merged in. Not responding and/or creating a way to merge/share ideas being worked on discourages future contributions... Some issues linger many months (e.g. #497) without any resolution which contribute to frustrate the people involved.

I suspect the waterline team is quite busy with Sails (v0.11 was just released) but a solution should be found for waterline to keep evolving and improving. Maybe envolve people not associated with sails?

PS: this is not to say I'm dissatisfied with waterline, it's a great ORM and the team has done a great job. Thanks!

randallmeeker:
I suggest sailsjs hack day! 1 day month where collaborators are actively online responding to pull requests and answering questions of contributors. Something organised so those of us can plan around.

devinivy:
I quite agree, and would be willing to help.  The number of PRs is getting a bit out of hand, and several of them pertain to bugs that are genuinely critical.  Some of them also pertain to features that are extremely welcomed.  I get the feeling some people who could be contributors are using their own forks right now.  I think that this is a great ORM too, and I really appreciate all the effort that has gone into the project!  At the same time the sheer number of PRs and outstanding issues is concerning.

dmarcelino:
@devinivy, when I started using waterline some of your comments helped me quite a bit. Namely the one about many-to-many through associations, which helped me implement through associations on my app, thanks!

It also makes me feel unsettled that there is no word if some sought after features will be implemented or when. To name a few:
* Transactions - #755
* Recursive populate - #256 / #308
* Many-to-many through associations release - #705
* Custom join table names - #497 / #725
* Composite indexes - #109

Some are more complex than others but it would be good to hear from the team about these and maybe see them being given milestones?

particlebanana:
@randallmeeker @devinivy @dmarcelino welcome to the Waterline core team! The only thing I ask is that any major commits you run by me first - shoot me an email. All features need to work across all our adapters.

This last piece (cross adapter) is why we don't have Transactions, Recursive populations, and composite indexes yet. Of the list above custom join table names, m-m through and composite are the easiest ones to get in anytime soon.

dmarcelino:
Thanks @particlebanana, I'll try to help where I can.

[...]

PR #896, opened 2015-03-23
Add adapter tests to waterline core...

dmarcelino:
... to ensure changes in waterline don't break adapters.

Back in issue #862 @particlebanana [said](https://github.com/balderdashy/waterline/issues/862#issuecomment-83643605):
> All features need to work across all our adapters.

Which I totally agree BUT currently the waterline core code doesn't **automatically** run against any adapters. As a quick way to find out if a waterline change breaks the adapters I suggest running waterline against the `sails-memory` integration tests on every build. In my machine this only makes the tests run for 2-3 seconds longer.

Another added benefit is that waterline core code coverage increases:

Coverage | Before | After
------------ | ------------- | -------------
Statements | 78.61% ( 2933/3731 ) | 84.4% ( 3149/3731 )
Branches | 67.87% ( 1335/1967 ) | 74.78% ( 1471/1967 )
Functions | 84.3% ( 521/618 ) | 89.64% ( 554/618 )
Lines | 83.12% ( 2778/3342 ) | 88.96% ( 2973/3342 )

Note: for now the adapter tests are always returning 0 because "Semantic Interface .find() should escape attribute names to prevent SQL injection attacks" test is broken in `sails-memory`. When this is properly fixed I'll change the integration test to return error.

Note2: We probably should also have some sort of automated build that would run tests agains every official adapter, maybe in the waterline-adapter-tests?

**EDIT:** changed "sails-memory" and "waterline-adapter-tests" dependencies to use the latest github versions. This is to cover the case where a developer is making intentional API changes to waterline-core. This way, as long as "sails-memory" and "waterline-adapter-tests" github versions are on pair the tests won't break.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
   },
   "devDependencies": {
     "mocha": "~2.1.0",
-    "should": "~4.4.0"
+    "sails-memory": "^0.10.3",
+    "should": "~4.4.0",
+    "waterline-adapter-tests": "^0.10.8"
   },
   "keywords": [
     "mvc",
@@ -47,7 +49,7 @@
   "repository": "git://github.com/balderdashy/waterline.git",
   "main": "./lib/waterline",
   "scripts": {
-    "test": "mocha test --recursive",
+    "test": "mocha test/integration test/structure test/support test/unit --recursive; node test/adapter/runner.js",
     "prepublish": "npm prune",
     "browserify": "rm -rf .dist && mkdir .dist && browserify lib/waterline.js -s Waterline | uglifyjs > .dist/waterline.min.js"
   },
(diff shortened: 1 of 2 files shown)


blah238:
+1 for this, but FYI your before/after numbers are the same (copy/paste mixup?)

devinivy:
Awesome :)

dmarcelino:
Thanks @blah238, fixed it now :wink:

dmarcelino:
@tjwebb, @randallmeeker and @particlebanana, would like to hear from you.

listepo:
+1

dmarcelino:
Thanks @tjwebb!

PR #898, opened 2015-03-25
Do not sync nested collections from Model#save.

devinivy:
Should fix critical bug #858.  After looking at more radical change in PR #861 to remove nested operation sync on update, @particlebanana suggested just targeting the issue from the model.  This passes all adapter tests that sails-memory currently passes using the waterline master branch and all vanilla waterline tests.

The approach is simply to remove attributes that are collections prior to passing the cloned model values to `ParentCollection.update`.

### DIFF
--- a/lib/waterline/model/lib/associationMethods/update.js
+++ b/lib/waterline/model/lib/associationMethods/update.js
@@ -37,8 +37,6 @@ var Update = module.exports = function(collection, proto, mutatedModels, cb) {
 
   // Clone values so they can be mutated
   var _values = _.cloneDeep(values);
-
-
   
   // For any nested model associations (objects not collection arrays) that were not changed,
   // lets set the value to just the foreign key so that an update query is not performed on the
@@ -56,7 +54,7 @@ var Update = module.exports = function(collection, proto, mutatedModels, cb) {
     }
   
     // If the key was changed, keep it expanded
-    if (mutatedModels.indexOf(key) !== -1) return;
+    if(mutatedModels.indexOf(key) !== -1) return;
 
     // Reduce it down to a foreign key value
     var vals = {};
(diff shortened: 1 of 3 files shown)


dmarcelino:
All `waterline-orientdb` tests pass when I run against this branch, which encourages me.

It would be good to see a test being added to this PR to see exactly the use case that this PR fixes (as well as to prevent future regressions).

devinivy:
@dmarcelino added!  With this PR, `Model#save` does not add, remove, or update records in a nested collection–you must use `add` and `remove` with ids or `Collection#update` directly.  The reason for this is that the nested update functionality is broken, and fixing it outright would be hellish.  Although it's not obvious from the test I wrote, here's the use-case that is broken:

```
Persons.findOne(666).populate('friends', {id: 42})
.then(function(person) {

    person.friends.remove(42);

    person.save(function() {

        /* now I have no friends,
         * since nested operations assumed
         * that this person had only one friend to begin with (friend 42),
         * which is not necessarily true.
         */
    });
});
```

dmarcelino:
@devinivy, I'm good with this :white_check_mark:
We probably want to make sure this change is mentioned on the release notes, because although it prevents a huge bug it may break the behaviour for some folks.

@particlebanana, @tjwebb, @rwoverdijk, @randallmeeker any concerns before we push the big green merge button?

devinivy:
I'd love for @particlebanana to peek at this since he was involved in the conversations around my first PR to fix this issue.  At the same time, this fix is time sensitive, as it is an issue of data integrity.

particlebanana:
Hey @devinivy this looks good. It worked by accident in the first place so this can just be a patch release. If you guys merge this stuff just shoot me an email and I can do an npm push today if it's ready. We have a good set of changes and it's been a bit since the last release.

devinivy:
I appreciate it, @particlebanana!

PR #903, opened 2015-03-27
#169: normalizes uppercase when sort passed as object

dmarcelino:
Issue #169.

Adds support for upper case `ASC` and `DESC` when sort criteria is passed as an object.


devinivy:
This looks good, and I'd love to get it through.  If the adapter tests are happy, I'm happy.

dmarcelino:
Thanks @devinivy! Yeap, it passes the adapter tests, the only annoying error is the dreaded *Semantic Interface .find() should escape attribute names to prevent SQL injection attacks*

PR #905, opened 2015-03-27
Fix .findOrCreateEach() test failures

dmarcelino:
This change aims to fix the the broken `.findOrCreateEach()` tests which are affecting most adapters.

Check balderdashy/waterline-adapter-tests#38 for more details.

### DIFF
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -88,11 +88,10 @@ module.exports = {
       // Check that each of the criteria keys match:
       // build a criteria query
       var criteria = {};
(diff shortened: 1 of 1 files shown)


tjwebb:
Looks good; for reference, can you describe what the issue was and how this fixes it?

tjwebb:
Ok I think this is sufficient: https://github.com/balderdashy/waterline-adapter-tests/issues/38#issuecomment-87098777

dmarcelino:
I was just about to copy & paste that here! :smile:

devinivy:
Awesome!  Figured I'd add, line 83 still has a hanging `var i = 0;`.  Not gonna do any harm, but might feel nice to nix that too.

PR #920, opened 2015-03-31
Refactor WLError, set stack writable

### LINKED ISSUE #913: WLError getter for "stack" breaks bluebird
https://github.com/balderdashy/waterline/blob/master/lib/waterline/error/WLError.js#L146-L148

see https://github.com/petkaantonov/bluebird/issues/552

Bluebird needs to be able to set the `.stack` property on `WLError` but is unable to do so

tjwebb:
- resolves #913 
- Includes some other minor cleanup

### DIFF
--- a/lib/waterline/error/WLError.js
+++ b/lib/waterline/error/WLError.js
@@ -1,11 +1,5 @@
-/**
- * Module dependencies
- */
-
 var util = require('util');
 var _ = require('lodash');
-
-
 
 /**
  * WLError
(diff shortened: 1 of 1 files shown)


devinivy:
Haven't tested this out, but it appears to be good!

tjwebb:
I'm using it locally now; the bug is gone, and I've noticed no other issues.

dmarcelino:
I'm good with this! :white_check_mark:

dmarcelino:
@tjwebb, it seems this fix is causing infinite loops when setting stack on WLError. Can you take  a look at #1001? Thanks.

PR #924, opened 2015-04-01
Cherry picked PR #829 (no conflicts)

blah238:
This is @nbiton's PR (Decreased processing time by making queries which involve population more efficient) https://github.com/balderdashy/waterline/pull/829 just without the package.json change.

### DIFF
--- a/lib/waterline/model/lib/internalMethods/defineAssociations.js
+++ b/lib/waterline/model/lib/internalMethods/defineAssociations.js
@@ -38,7 +38,7 @@ var Define = module.exports = function(context, proto) {
     value: {}
   });
 
-  var attributes = _.cloneDeep(context._attributes) || {};
+  var attributes = context._attributes || {};
   var collections = this.collectionKeys(attributes);
   var models = this.modelKeys(attributes);
 
@@ -69,7 +69,7 @@ Define.prototype.collectionKeys = function(attributes) {
   // Find any collection keys
   for(var attribute in attributes) {
     if(!hasOwnProperty(attributes[attribute], 'collection')) continue;
-    collections.push(attribute);
+    collections.push(_.cloneDeep(attribute));
   }
 
   return collections;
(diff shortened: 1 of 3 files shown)


devinivy:
If this passes the adapter tests, it's good!  Looks great.  As soon as someone confirms that, merge away!  I'd like to see this in a release by the end of the week with #889 if possible.  @particlebanana is that reasonable?

nbiton:
I had planned to make another attempt at improving on this by changing it so these lines:
https://github.com/nbiton/waterline/blob/316c3cff79103279ff2268132994f0752243551e/lib/waterline/model/lib/model.js#L39-L60
will only be executed once per model definition, during the application initialization phase, instead of having it being executed once for every record pulled out of the DB.

It looks to me that if we keep a map object for **schema -> record base** which will be populated during the initialization phase, and then will be used in the model constructor to associate the *record base* to the record object as a prototype, we can achieve that.
This possible solution will also work around the issue which @devinivy mentioned in https://github.com/balderdashy/waterline/pull/829#issuecomment-74443706

I'm hoping to find the time to try this myself, soon.

devinivy:
:+1: that sounds nice

particlebanana:
@devinivy traveling a bit today but I'll get these published before the weekend. I want to dig into both of them and make sure I understand all the changes. You guys are kicking ass and I don't want to be a bottleneck on getting stuff pushed out so I'll make it quick!

blah238:
@particlebanana Sounds great, will there also be a corresponding Sails release? :pray:

blah238:
@nbiton I think you are definitely on to something there.

I tried commenting out [these lines](https://github.com/nbiton/waterline/blob/316c3cff79103279ff2268132994f0752243551e/lib/waterline/model/lib/model.js#L51-L53) and again halved the response time and quartered (!) the memory consumption of the scenario described in [this post](https://github.com/balderdashy/sails-postgresql/issues/141#issue-64400447).

Those few lines seem to be **extremely** expensive, and the results are only used [during saves](https://github.com/nbiton/waterline/blob/316c3cff79103279ff2268132994f0752243551e/lib/waterline/model/lib/defaultMethods/save.js#L46-L65), so if we can figure out a way to not avoid doing that on every find for every record that would be another huge optimization win, IMO.

CC @particlebanana @devinivy

blah238:
Thanks @particlebanana ! And @nbiton of course for the actual patch!

I will probably open a separate issue regarding the `associationsCache` stuff after I've had a chance to get some more metrics.

devinivy:
@blah238 good finds.  Do you mind starting an issue with a list of places in the code that could use a performance makeover?  You have some good hunches that we should look into.

Issue #783, opened 2015-01-02
destroy many to many relationship with custom primaryKey in both tables


camel-chased:
```javascript
//model user
schema        : true,
	autoPK        : false,
	autoCreatedAt : false,
	autoUpdatedAt : false,
attributes:{
    id:{
        primaryKey:true,
        columnName:'u_id'
    },
    houses:{
        collection:'house',
        via:'users'
    }
}

//model house
schema        : true,
	autoPK        : false,
	autoCreatedAt : false,
	autoUpdatedAt : false,
attributes:{
    id:{
        primaryKey:true,
        columnName:'h_id',
    },
    users:{
        collection:'user',
       via:'houses'
    }
}
```

try to destroy this with some houses/users in db (mysql)
ER_BAD_FIELD_ERROR: Unknown column 'undefined' in 'where clause'
when i change columnName to 'id' in both tables, everything goes nice

esopian:
Same issue here.  Would love to find a solution for this.  Here's the mysql call that waterline tries to make.

```
SELECT `restaurantcuisine`.`ID`, `restaurantcuisine`.`RESTAURANT_ID`, `restaurantcuisine`.`CUISINE_ID` FROM `RESTAURANT_CUISINE` AS `restaurantcuisine`  WHERE `restaurantcuisine`.`RESTAURANT_ID` IN (undefined)  ORDER BY `restaurantcuisine`.`ID` DESC
```

dmarcelino:
@camel-chased, can you try the fix in #928 to check if it fixes your issue please?

camel-chased:
i don't work on waterline any more but i make something like to make it work: 
https://github.com/camel-chased/sails-redis/commit/53d7f4e584c9e40a210265b6e8c5d0368afe156d

dmarcelino:
Thanks

dmarcelino:
Fixed via #1022. A big thanks to @particlebanana! :wink:

PR #928, opened 2015-04-03
fixing many to many deletes

esopian:
Destroy will fail to create a valid sql query in many to many relationships with custom primary key column names.  The resulting query will include undefined columns

```
IN (undefined, undefined, undefined)
```

### DIFF
--- a/lib/waterline/query/dql/destroy.js
+++ b/lib/waterline/query/dql/destroy.js
@@ -74,6 +74,8 @@ module.exports = function(criteria, cb) {
         break;
       }
 
+      var transformedValues = _.map(result, self._transformer.unserialize.bind(self._transformer))
+
       function destroyJoinTableRecords(item, next) {
         var collection = self.waterline.collections[item];
         var refKey;
@@ -88,7 +90,14 @@ module.exports = function(criteria, cb) {
         // than crashing.
         if(!refKey) return next();
 
-        var mappedValues = result.map(function(vals) { return vals[pk]; });
+        // Make sure we don't return any undefined pks
+        var mappedValues = transformedValues.reduce(function(memo, vals) {
+          if (vals[pk]) { 
+            memo.push(vals[pk]); 
+          }
+          return memo; 
+        }, []);
+
         var criteria = {};
 
         if(mappedValues.length > 0) {
--- a/lib/waterline/query/dql/destroy.js
+++ b/lib/waterline/query/dql/destroy.js
@@ -92,7 +92,7 @@ module.exports = function(criteria, cb) {
 
         // Make sure we don't return any undefined pks
         var mappedValues = transformedValues.reduce(function(memo, vals) {
-          if (vals[pk]) { 
+          if (vals[pk] !== undefined) {
             memo.push(vals[pk]); 
           }
           return memo; 
--- a/test/integration/model/association.destroy.manyToMany.js
+++ b/test/integration/model/association.destroy.manyToMany.js
@@ -1,0 +1,112 @@
+var Waterline = require('../../../lib/waterline'),
+    _ = require('lodash'),
+    assert = require('assert');
+
+describe('Model', function() {
+  describe('associations Many To Many', function() {
+    describe('.destroy()', function() {
+
+      /////////////////////////////////////////////////////
+      // TEST SETUP
+      ////////////////////////////////////////////////////
+
+      var collections = {};
+      var prefDestroyCall;
+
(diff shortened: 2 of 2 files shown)


devinivy:
Looks good to me.  Does there happen to be a related issue for this?  Also, can someone verify that the added test will fail without this change?  Lastly, we need to test this against the adapter tests.  Thanks, @esopian !

esopian:
No related issue that I know of.  We had to patch for behavior we saw in our project.
We failed the test first before fixing but someone should absolutely double check me on this  :)

lib/waterline/query/dql/destroy.js
```diff
@@ -88,7 +90,14 @@ module.exports = function(criteria, cb) {
         // than crashing.
         if(!refKey) return next();
 
-        var mappedValues = result.map(function(vals) { return vals[pk]; });
+        // Make sure we don't return any undefined pks
+        var mappedValues = transformedValues.reduce(function(memo, vals) {
+          if (vals[pk]) {
```
dmarcelino: What if `vals[pk] === 0`? That would be a valid pk. I think we should replace this line for:
```javascript
 if (vals[pk] !== undefined) { 
```

esopian: Agreed.  pushed a fix

dmarcelino:
I wonder if this can fix issue #783.

esopian:
It should fix #783 as well.

lib/waterline/query/dql/destroy.js
```diff
@@ -74,6 +74,8 @@ module.exports = function(criteria, cb) {
         break;
```
dmarcelino: Any ideas why the adapter returns `undefined` PKs in `result`?

We probably need this change to sanitise adapters output but it looks like the root cause of this issue may be caused by the adapters themselves.

lib/waterline/query/dql/destroy.js
```diff
@@ -74,6 +74,8 @@ module.exports = function(criteria, cb) {
...
       }
 
+      var transformedValues = _.map(result, self._transformer.unserialize.bind(self._transformer))
```
dmarcelino: Any ideas why the adapter returns `undefined` PKs in `result`?

We probably need this change to sanitise adapters output but it looks like the root cause of this issue may be caused by the adapters themselves.

devinivy:
@dmarcelino good point... we really need to look deeper into this issue.

devinivy:
Got to the bottom of this!  The results are coming directly from the adapter, so they haven't been transformed.  That means that results are keyed by column name.  Will submit a PR.  Unclear if this one should live on, or if it could mask other issues.  I think in this sort of situation, waterline should ideally have thrown an error early on.

dmarcelino:
Good catch @devinivy! That seems to be the best solution. Assuming your PR will fix the problem, we won't need this fix, since it can mask other issues (as you said), though we should keep the tests which are most welcomed. :)

lib/waterline/query/dql/destroy.js
devinivy: This makes sense.  The adapters pass back values keyed with column names.  It's the transformer's duty to map those column names to waterline attribute names.  Since the raw adapter method was used [here](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/destroy.js#L55), but [this block of code](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/destroy.js#L66-L75) finds an _attribute_ name, that's why the problem arises.

devinivy:
@esopian does have the correct fix and the tests are great, so I may not have to make a PR.  Any feelings about,

1. Replacing the `undefined` filter with an assertion that will error and, 
2. Possibly replacing the full transformation with a column name lookup rather than an attribute name lookup [here](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/destroy.js#L66-L75).

dmarcelino:
1. if the tests still pass I think that makes sense!

dmarcelino:
@particlebanana, @tjwebb, what do you guys think?

dmarcelino:
@esopian, any chance you could follow up on @devinivy's suggestions in https://github.com/balderdashy/waterline/pull/928#issuecomment-93990300? Thanks!

particlebanana:
Submitted #1022 that expands on this and makes it so we don't need to use the transformer.

dmarcelino:
@particlebanana's solution does seem simpler.

PR #930, opened 2015-04-05
Fix crash for nested model create within update query

mkeremguc:
add failing test and try fix for #533 and possibly #644

### DIFF
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -137,9 +137,7 @@ function createBelongsTo(valuesObject, cb) {
       var pk = val[model.primaryKey];
 
       valuesObject.values[item] = pk;
-      
-      // now we have pk value attached, remove it from models
-      delete valuesObject.associations.models[item];
+      _.remove(valuesObject.associations.models, function(_item) { return _item == item; });
       next();
     });
 
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -137,6 +137,8 @@ function createBelongsTo(valuesObject, cb) {
       var pk = val[model.primaryKey];
 
       valuesObject.values[item] = pk;
+      
+      // now we have pk value attached, remove it from models
       _.remove(valuesObject.associations.models, function(_item) { return _item == item; });
       next();
     });
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -137,7 +137,9 @@ function createBelongsTo(valuesObject, cb) {
       var pk = val[model.primaryKey];
 
       valuesObject.values[item] = pk;
-      _.remove(valuesObject.associations.models, function(_item) { return _item == item; });
+      
+      // now we have pk value attached, remove it from models
+      delete valuesObject.associations.models[item];
       next();
     });
 
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -137,6 +137,7 @@ function createBelongsTo(valuesObject, cb) {
       var pk = val[model.primaryKey];
 
       valuesObject.values[item] = pk;
+      _.remove(valuesObject.associations.models, function(_item) { return _item == item; });
       next();
     });
 
--- a/test/unit/query/query.update.nested.js
+++ b/test/unit/query/query.update.nested.js
@@ -73,6 +73,81 @@ describe('Collection Query', function() {
         });
       });
     });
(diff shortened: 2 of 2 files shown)


dmarcelino:
So, from what I can gather, the code changes the output of `query/dql/update.js#createBelongsTo()`. Using the test example, instead of returning:
```javascript
valuesObject.associations = { collections: [], models: [ 'nestedModel' ] }
```
it returns:
```javascript
valuesObject.associations = { collections: [], models: [] }
```
Not sure if this can have any other unintended consequences but it does passes all tests.

@particlebanana, @tjwebb, @devinivy, can you please take a look at this change? I confirmed the submitted test fails and then it passes after the change.

mkeremguc:
`createBelongsTo()` creates the nested model. however, it also calls `updateRecords()` (line 52). so either it should not try to update the newly created record or it should just remove the new-created-model from `models` array (as in my patch) so it doesn't try to update and crash the whole app.

lib/waterline/query/dql/update.js
```diff
@@ -137,6 +137,7 @@ function createBelongsTo(valuesObject, cb) {
       var pk = val[model.primaryKey];
 
       valuesObject.values[item] = pk;
+      _.remove(valuesObject.associations.models, function(_item) { return _item == item; });
```
devinivy: Perhaps this could be simpler and have a comment?
```javascript
// now we have pk value attached, remove it from models
delete valuesObject.associations.models[item];
```

devinivy:
This seems like it may be a reasonable patch for optimization ("don't update a record that was just created"), but it doesn't seem to solve the underlying problem at hand.  You had a very specific problem in nestedUpdates/update.js (see https://github.com/balderdashy/waterline/issues/533#issuecomment-68065865):
>then within async mapAssociations /nestedOperations/update.js line 516:
```
>var model = self.waterline.collections[association];
```
>above, "association" is equal to "photo" (coming from async.each above) and "model" goes undefined. because there is no such model named "photo"...

This, to me, is the issue at hand.  Somewhere something is set incorrectly, which conflates "photo" (an attribute name?) with a collection name.  That causes the context to be completely broken (set to `global`) during `createRecord`.  I don't think your test recreates this particular issue.

This [line](https://github.com/mkeremguc/waterline/blob/fix-crash-nestedCreate-on-update/lib/waterline/utils/nestedOperations/update.js#L513) may be wrong, however:
```javascript
var model = self.waterline.collections[association];
```
Because an association name here is an _attribute's_ name rather than a collection's name.  You want to use the `references` property from the schema as shown [here](https://github.com/mkeremguc/waterline/blob/fix-crash-nestedCreate-on-update/lib/waterline/utils/nestedOperations/update.js#L94).

I expect that [these adapter tests](https://github.com/balderdashy/waterline-adapter-tests/blob/master/interfaces/associations/belongsTo/update.nested.js) pass because the collection name and association name are identical.

Yeesh! @dmarcelino can you check that out if you have a chance?

mkeremguc:
@devinivy good point there. that is why i always doubted that it was a valid fix since i didn't have the time to explore all the magic goes in waterline back there. also lets update all the tests with attr names to be different than model names..

devinivy:
How does this work for you? https://github.com/devinivy/waterline/commit/0e307c59515c70c11f065965c7156e3ea66361ca

devinivy:
PR for adjusted adapter tests: https://github.com/balderdashy/waterline-adapter-tests/pull/44
PR for fix: https://github.com/balderdashy/waterline/pull/934

mkeremguc:
@devinivy,  just tested your commit and there is a problem there. it creates the nested model twice

devinivy:
Thanks for testing.  Amazingly enough, I expect our PRs may need to come together to fix this issue!  The code I fixed allowed a new bug to appear, which I bet this PR resolves.

lib/waterline/query/dql/update.js
mkeremguc: done

mkeremguc:
ok good, however your suggestion as `delete valuesObject.associations.models[item];` won work causes .models is array of string so i reverted the commit of your suggestion

devinivy:
Ah, sorry for misleading!

lib/waterline/query/dql/update.js
mkeremguc: `delete valuesObject.associations.models[item];` won't work causes .models is array of string

devinivy:
See comments on #934: these are ready to be merged together.

PR #932, opened 2015-04-06
FindoOrCreate arrays of object fix

Globegitter:
This is as discussed in https://github.com/balderdashy/waterline-adapter-tests/issues/38

For some reason the unit-test I created did not catch the bug that broke my app. This now fully supports the use case I am using and that is described in https://github.com/balderdashy/waterline-docs/blob/master/query-methods.md#findorcreate-search-criteria-values-callback-, which is passing arrays of objects.

I tested now against https://github.com/balderdashy/waterline-adapter-tests and with `sails-memory` only get 1 failing test that is unrelated to my change.

@dmarcelino could you review?

### DIFF
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -84,14 +84,20 @@ module.exports = {
 
     async.eachSeries(valuesList, function (values, cb) {
       if (!_.isObject(values)) return cb(new Error('findOrCreateEach: Unexpected value in valuesList.'));
-
+      console.log('eachSeries iterator', i);
+      console.log('eachSeries iterator', attributesToCheck[i]);
       // Check that each of the criteria keys match:
       // build a criteria query
       var criteria = {};
-      
-      attributesToCheck.forEach(function (attrName) {
+
+      Object.keys(attributesToCheck[i]).forEach(function (attrName, i) {
+        console.log('attrName', attrName);
         criteria[attrName] = values[attrName];
       });
+
+      console.log('criteria', criteria);
+      console.log('values', values);
+      i++;
 
       return self.findOrCreate.call(self, criteria, values, function (err, model) {
         if(err) return cb(err);
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -84,20 +84,20 @@ module.exports = {
 
     async.eachSeries(valuesList, function (values, cb) {
       if (!_.isObject(values)) return cb(new Error('findOrCreateEach: Unexpected value in valuesList.'));
-      console.log('eachSeries iterator', i);
-      console.log('eachSeries iterator', attributesToCheck[i]);
       // Check that each of the criteria keys match:
       // build a criteria query
       var criteria = {};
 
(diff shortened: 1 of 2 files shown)


devinivy:
I'm uneasy about this.  Can you effectively "document" how you're using `findOrCreateEach` to make sure I understand?  This includes support for a departure from how it is currently used, while we are trying to deprecate the method altogether.  Additionally, in your usage of the method, it seems as though it is a requisite that the criteria and values arrays be the same length–so this would need to create an error if that were not the case.  There are also breaking cases when mixing object and non-object values in the criteria list.  My initial reaction is that you should loop `findOrCreate` within your application.

Globegitter:
@devinivy you are right, will add an error for that.

I am currently using it as described here: https://github.com/balderdashy/waterline/blob/master/test/unit/query/query.findOrCreate.js#L61

But strangely enough this unit test passes, even though it shows the same issues as my app has. I can console.log that the `forEach` loop creates bogus but for some reason the model has already been created at this point with the timestamps etc.

devinivy:
Alright!  I'd just like the function signature and accepted values to be reflected more transparently in the code, with violations of that signature provided useful errors.

Globegitter:
@devinivy so should the function signature match the one of `findOrCreate` here https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/composite.js#L24 ?

devinivy:
I guess I'm thinking of either,

1. The first argument and the second argument are arrays of the same length, respectively criteria and values paired by array index.  Criteria items and value items are both objects.
2. The first argument is an array of attribute values, and the second argument is an array of values.  Criteria for each value is built by picking the attributes (specified in the first argument) off of the values object.

In the first case the arguments are two arrays of objects, both the same length.  In the second case, the first argument is an array of strings and the second is an array of objects.

If I have that wrong, let me know!

Globegitter:
What you said made entire sense, but apparently according to this test https://github.com/balderdashy/waterline/blob/master/test/unit/query/query.findOrCreateEach.js#L91this is not true and the arrays of objects can indeed have different length.

I am not really sure how this behaviour is intended. Intuitively I would say if it does not find `name: foo` it creates both. Right now I will work on it keeping backwards compatibility to not just break things.

The other way around I suppose you could be searching for multiple models and if one of that does not exist you want to create the one element you passed. Does that make sense?

Globegitter:
But thinking more about it that leads to really weird edge-case behaviour. I honestly think the test I posted above makes little sense. But I suppose we could have an edge-case scenario if either of the arrays are or length 1.

Globegitter:
Ok I now added edge-case behaviour for the current failing test, if `attributesToCheck.length === 1`, but I have not added an edge-case behaviour if `valuesList.length === 1` since there is no test for it and I see no point in adding unnecessary complexity for that.

Let me know if you want me to implement it any different way.

Globegitter:
So, all current tests are passing plus I added more tests here https://github.com/balderdashy/waterline-adapter-tests/pull/45 that are currently failing but passing with this PR.

dmarcelino:
I'm good with both this PR and balderdashy/waterline-adapter-tests#45 :white_check_mark:

devinivy:
:heavy_check_mark:

dmarcelino:
Thanks @Globegitter, @devinivy!

PR #934, opened 2015-04-07
Use correct foreign collection during nested update creation/link step.

devinivy:
Ref #533 and #930.

I think this is the correct fix for the problem referenced in issues above.  This should resolve the bugs caught in the adjusted adapter tests (https://github.com/balderdashy/waterline-adapter-tests/pull/44).

I'd call this fairly high priority for evaluating.  Happy to answer questions about the problem at hand and this solution.

### DIFF
--- a/lib/waterline/utils/nestedOperations/update.js
+++ b/lib/waterline/utils/nestedOperations/update.js
@@ -510,7 +510,12 @@ function createNewRecords(parent, recordsToCreate, cb) {
   // For each association, run the createRecords function
   // in the model context.
   function mapAssociations(association, next) {
-    var model = self.waterline.collections[association];
+    
+    // First, pull the model attribute's referenced (foreign) collection
+    var attribute = self.waterline.schema[self.identity].attributes[association];
+    var referencedCollection = attribute.references;
+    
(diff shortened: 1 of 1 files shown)


devinivy:
It's likely the case that #930 and this PR will have to come together to fix the issue.  According to @mkeremguc, two nested models are being plopped in his DB.  I haven't verified, but I believe it: one from the `createBelongsTo` routine in dql/update, and one from update's nested operations.

mkeremguc:
thanks for indicating the high priority finally :)

tjwebb:
Yes currently this is actually **top** priority.

particlebanana:
Looks good - exposes a bug that creates duplicate records but #930 takes care of that. Lets merge these in and give it a good test run to see if anything else stands out.

devinivy:
Awesome.  I'd love it if someone merged #930 and #934 into WL then tested against waterline-adapter-tests with https://github.com/balderdashy/waterline-adapter-tests/pull/44 merged in.  See what we get!

dmarcelino:
@devinivy, just did that and all tests passed: :white_check_mark:

PR #935, opened 2015-04-07
Add eslint and improve test-runner

Globegitter:
This PR is supposed to get a discussion started based on success we had with [eslint](http://eslint.org/) over at Sane Stack. I added [eslint](http://eslint.org/) via [mocha-eslint](https://github.com/BadgeLabs/mocha-eslint) so waterline can have consistent code-styling plus it also catches small programmers' mistakes such as undefined variables, mismatching brackets etc.

I have just added the same `.eslintrc` file that we have over at sane and that may or may not be exactly suitable for sailsjs/waterline so @particlebanana @mikermcneil  @dmarcelino @devinivy @tjwebb let me know what you think of adding that and which rules you would like to have changed if any. For all the possible rules see: http://eslint.org/docs/rules/

If you run this locally you will see that the current code breaks a lot of the rules. So once we have settled on which rules to use I would be happy to fix them.

Furthermore I moved everything from the `Makefile` to `test/runner.js` as well as `test/coverage.js` to make things a bit more JS programmer friendly.

I am not quite happy about the `test/coverage.js` file so will improve that further once I familiarise myself more with istanbul.


Makefile
```diff
@@ -1,24 +0,0 @@
...
-test: test-unit test-integration
-
-test-unit:
-	@echo "\nRunning unit tests..."
-	@NODE_ENV=test mocha test/integration test/structure test/support test/unit --recursive
-
-test-integration:
-	@echo "\nRunning integration tests..."
-	rm -rf node_modules/waterline-adapter-tests/node_modules/waterline;
-	ln -s $(ROOT) node_modules/waterline-adapter-tests/node_modules/waterline;
```
dmarcelino: From looking at `test/adapter/runner.js` it's not clear to me how you ensure that the integration tests run against waterline **edge**. `waterline-adapter-tests` will have as dependency waterline **stable** and this step ensures we are testing against any changes done in waterline.

dmarcelino:
My first impressions:
* This may be better divided into 2 PRs since limiting and getting rid of the Makefile will probably generate 2 different sets of comments (but lets see how it goes);
* Makefile:
  * Coverage, you've already brought this up but just in case, it usually needs to be run with **_mocha**, are you getting the same/similar results as in here: https://github.com/balderdashy/waterline/pull/896#issue-63777327?
* Linting:
  * Applied to the source: :+1: , assuming it's not too late for waterline :smiley: 
  * Applied to the tests: I think this is being a bit too pedantic and it's a disincentive for (some) people to write more tests. I prefer to have more tests (even breaking linting) than less tests.
  * I'm more accustomed to jshint, what are the differences compared to eslint?

test/runner.js
```diff
@@ -0,0 +1,68 @@
...
+  var Mocha = require('mocha');
+
+  //Define Mocha
+  var mocha = new Mocha({
+    // For some reason, tests take a long time on Windows (or at least AppVeyor)
+    timeout: (process.platform === 'win32') ? 30000 : 18000,
+    reporter: 'spec'
+  });
+
+  root = 'test/unit';
```
dmarcelino: What about `test/integration`, `test/structure` and `test/support`? These are currently being ran as well.

Globegitter:
@dmarcelino You are right about the folders, was on oversight from me, mostly because the `test-integration` naming is imo a bit confusing. Since now we have the integration adapter tests and the normal integration tests. So I renamed the adapter tests just to `adapter`.

So when you run `npm test` it runs all tests, but you can specify `npm test -- unit` or `npm test -- adapter` to either just run the adapter or normal tests. Furthermore you run `npm run-script coverage` for the coverage and I am getting back:
```
=============================== Coverage summary ===============================
Statements   : 82.15% ( 3074/3742 )
Branches     : 72.23% ( 1433/1984 )
Functions    : 86.63% ( 538/621 )
Lines        : 86.76% ( 2909/3353 )
================================================================================
```

Which is pretty close to what you posted. I think it works the way I specified it because the `runner.js` calls Mocha. But I will look more into it, since I have no experience with Istanbul.

Yep, eslint is quite comparable to jshint, but the main differences, taken from their Readme are:

- ESLint uses Espree for JavaScript parsing.
- ESLint uses an AST to evaluate patterns in code.
- ESLint is completely pluggable, every single rule is a plugin and you can add more at runtime.

In practical terms that means that ESLint does more work than JSHint and is a bit more slower, but it actually checks that your code is runnable and also allows certain checks that JSHint could not do (e.g. complexity check: http://eslint.org/docs/rules/complexity or that you can not nest callbacks more than x times, etc). More importantly though it supports ES6/7 JS much better and has been written with a plugin-system from core, so you can piece together/change pretty much everything you want.

dmarcelino:
@Globegitter, the test naming is mostly my fault, I didn't want to rename the existing integration folder and yet I called the tests integration tests. Adapter tests sounds good.

Thanks for clarifying about ESLint.

Generally speaking I'm OK with the changes, I would just like that we don't enforce linting on the tests themselves to avoid any hurdles that can keep people from writing more tests.

Globegitter:
@dmarcelino I personally like having a clean code-base throughout but if contributors are for removing it for the `test` folder, I of course respect that decision. 

Another more middle-ground option would be to have a separate `.eslintrc` in the `test` folder and just put everything as a warning, which means linting would not be enforced but the developers who care at least get to see what's up with the code and could fix it.

IanVS:
@dmarcelino, This blog post introducing ESLint back in 2013 is a good summary of its differences and improvements over JSHint: http://www.nczonline.net/blog/2013/07/16/introducing-eslint/

@Globegitter already covered many of the points, but the link above gives some good background.

mphasize:
Full support for getting this one in. I believe Waterline and Sails could benefit from a clear code style. Having eslint generate warnings for errors in the tests sounds like the right approach.

dmarcelino:
Thanks @IanVS.

During a transition stage both eslint tasks will need to be warnings until we clean all errors... Unless someone wants to take that herculean task upon himself! Any volunteers? :smile: 

BTW @Globegitter, I would prefer for linting to happen in package.json `pretest` script. That way it will run before any test and it also allows to lint by running `npm run pretest`. In sails-orientdb we have:
```
"pretest": "./node_modules/.bin/jshint ./lib",
```

In reality we also have `"lint": "./node_modules/.bin/jshint ./lib"` so people can do `npm run lint` but that's optional :)

Globegitter:
Yep @dmarcelino that's what we switched to recently as well. Will update PR asap.

dmarcelino:
@Globegitter, given PR #1070 (Add eslint for source code) has been merged, can you possibly remove the eslint changes from this one and keep the makefile ones? Thanks!

IanVS:
As an aside, @dmarcelino, you don't need to specify `node_modules/.bin` in npm scripts, as that directory is automatically added to PATH when `npm run` is used.  

So instead of :

```
"pretest": "./node_modules/.bin/jshint ./lib",
```

You could use:
```
"pretest": "jshint ./lib",
```

Or even:
```
"lint": "jshint ./lib", 
"pretest": "npm run lint"
```

particlebanana:
We have `eslint` now. As this is in a conflict any changes to the test runner will need to be submitted as a new PR. Closing to clean up some stale PR's.

Issue #931, opened 2015-04-06
Future of findOrCreateEach and dynamic finders


dmarcelino:
According to PR #843 `findOrCreateEach` is deprecated:

@tjwebb 
> Personally I'd like to deprecate (and remove) all dynamic finders, since it's just senseless proliferation of API surface area.

@particlebanana:
> I would love to remove the dynamic finders. Whenever we feel like an 0.11 release is ready we can take a pass and remove all that stuff.

The questions are:
* Should we strip the `findOrCreateEach` from waterline on `0.11`?
* Should we strip all dynamic finders code from waterline on `0.11`?
* Should we move dynamic finders to another project?
* Any other thoughts regarding dynamic finders?

cc @devinivy, @randallmeeker, @Globegitter, @RWOverdijk

Globegitter:
The way I see it is that `findOrCreateEach` indeed should be deprecated and `findOrCreate` should support both use-cases. Which it is pretty close to supporting right now anyway.

devinivy:
I don't care if we strip `findOrCreateEach` or dynamic finders out of the API.  But I'm curious how widely the dynamic finders are used.  It is a pretty typical feature of ORMs, I think.

If we do remove `findOrCreateEach`, we should stop using it internally to waterline core.  This was brought-up with @Globegitter [here](https://github.com/balderdashy/waterline-adapter-tests/issues/38#issuecomment-90078230).

devinivy:
Also, familiarize yourself with what `findOrCreateEach` actually does if you don't already know!  It may not be as obvious as one would expect: https://github.com/balderdashy/waterline-adapter-tests/blob/master/interfaces/semantic/findOrCreateEach.test.js#L29

TLDR: 1st argument is an array of attributes to `findOne` using array of objects in 2nd argument.

Globegitter:
Also on that note, I am halfway-through implementing a `findAndModify` function with upsert support. See: http://docs.mongodb.org/manual/reference/command/findAndModify/

For that I also followed the current way and implemented a `findAndModifyEach`, but purely intended as a helper function. I don't want to drift off the main topic of this discussion but I think it fits here and would be a good questions if we want this functionality (I think we should have these slightly more advanced functions. You don't need them very often, but once you do need them they are tedious enough to implement) and how to implement all use-cases without `findAndModifyEach`.

Also does someone have a list of what all the dynamic finders are?

dmarcelino:
helper methods:

  - createEach
[...]

PR #939, opened 2015-04-08
FindAndModify + upsert

Globegitter:
I know there is the current discussion going on about `findOrCreateEach` (
https://github.com/balderdashy/waterline/issues/931) so it might not have been the wisest choice to add a `findAndModifyEach` but I needed a `findAndModify` functionality (see http://docs.mongodb.org/manual/reference/method/db.collection.findAndModify/) and it seemed like a good solution to put it such a way that other people can use it as well.

Usage is `Model.findAndModify(search critieras, values, [options, cb]);`

where-as default options look like that:

```js
{
  upsert: false, // on true automatically inserts the values if nothing is found
  new: true, // if false returns the model before it was updated/created
  mergeArrays: false // quite a special case that I needed and I am still testing in my app. When any of the values given contain an array it is merged to the found object rather than just replaced.
}
```

I added passing unit tests as well as adapter-tests, see https://github.com/balderdashy/waterline-adapter-tests/pull/47
There is something strange going on, even though the `findAndModify` passes the unit tests and works  in my app (via `npm i git+https://github.com/Globegitter/waterline.git#findAndModify`) the adapter-tests do not pass and throw the same error throughout:

```js
9) Semantic Interface .findAndModfiy() should return a model instance:
     TypeError: undefined is not a function
      at Context.<anonymous> (Git/waterline/node_modules/waterline-adapter-tests/interfaces/semantic/findAndModify.test.js:73:20)
      at Test.Runnable.run (Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runnable.js:194:15)
      at Runner.runTest (Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:355:10)
      at Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:401:12
      at next (Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:281:14)
      at Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:290:7
      at next (Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:234:23)
      at Immediate._onImmediate (Git/waterline/node_modules/waterline-adapter-tests/node_modules/mocha/lib/runner.js:258:5)
      at processImmediate [as _immediateCallback] (timers.js:361:17)
```

I have no idea why this happens or how to fix this. So apart from if this actually gets merged, anyone got any idea of why that could be?

Also it does not work completely in my app via `sails-postgresql`, I seem to be getting this error. I am not sure if that is related to my code or to `sails-postgresql`.

```
db_1     | ERROR:  syntax error at or near "LIMIT" at character 140
db_1     | STATEMENT:  UPDATE "subject" AS "subject" SET "researchIds" = $1, "text" = $2, "createdAt" = $3, "updatedAt" = $4  WHERE LOWER("subject"."text") = $5  LIMIT 1 RETURNING *
````

### DIFF
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -127,6 +127,76 @@ module.exports = {
       if(err) return cb(err);
       cb(null, models);
     });
+  },
+
+  // If an optimized findAndModify exists, use it, otherwise use an asynchronous loop with create()
+findAndModifyEach: function(attributesToCheck, valuesList, options, cb) {
+  console.log('In findAndModifyEach in aggregateQueries.js');
+  var self = this;
+  var connName;
+  var adapter;
+
+  if(typeof options === 'function') {
+    cb = options;
+    // set default values
+    options = {
+      upsert: false,
+      new: false,
+      mergeArrays: false
+    };
   }
+
+  // Normalize Arguments
+  cb = normalize.callback(cb);
+
+  // Clone sensitive data
+  attributesToCheck = _.clone(attributesToCheck);
+  valuesList = _.clone(valuesList);
+
+  // Custom user adapter behavior
+  if(hasOwnProperty(this.dictionary, 'findAndModifyEach')) {
+    connName = this.dictionary.findAndModifyEach;
+    adapter = this.connections[connName]._adapter;
+
+    if(hasOwnProperty(adapter, 'findAndModifyEach')) {
+      return adapter.findAndModifyEach(connName, this.collection, valuesList, options, cb);
+    }
+  }
+
+  // Build a list of models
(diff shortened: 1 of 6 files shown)


dmarcelino:
@Globegitter, I think the chances of this being merged to master are slim, you will probably get a more favourable audience if you PR this against `0.11` branch (still depending on the outcome of balderdashy/waterline#931). The idea is that 0.10.x is being used to fix bugs and new features are to be added to 0.11.

Globegitter:
Ok will do.

PR #941, opened 2015-04-08
897: if the adapter tests fail return code 1 so the automated build fails

dmarcelino:
This is to avoid the case where the tests are broken but the build passes.

Part of issue #897.


devinivy:
:heavy_check_mark:

dmarcelino:
Thanks @devinivy!

PR #942, opened 2015-04-08
Make shallow copy of criteria array for findOne

Globegitter:
Again, fixing a bug as discussed with @devinivy and @dmarcelino .

@devinivy I tried adding the shallow copy at the place you suggested first: https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/basic.js#L35, but that did haven no affect in my code. Adding it here fixes the issue!

### DIFF
--- a/lib/waterline/adapter/dql.js
+++ b/lib/waterline/adapter/dql.js
@@ -132,7 +132,12 @@ module.exports = {
    * @return {[type]}            [description]
    */
   findOne: function(criteria, cb) {
-
+    
+    //make shallow copy of criteria so original does not get modified
+    if (Array.isArray(criteria) || _.isObject(criteria)) {
+      criteria = _.clone(criteria);
+    }
+    
     // Normalize Arguments
     cb = normalize.callback(cb);
 
--- a/lib/waterline/adapter/dql.js
+++ b/lib/waterline/adapter/dql.js
@@ -5,7 +5,6 @@
 var normalize = require('../utils/normalize');
 var schema = require('../utils/schema');
 var hasOwnProperty = require('../utils/helpers').object.hasOwnProperty;
-var _ = require('lodash');
 
 
 /**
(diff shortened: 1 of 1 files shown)


lib/waterline/adapter/dql.js
```diff
@@ -132,7 +131,12 @@ module.exports = {
    * @return {[type]}            [description]
    */
   findOne: function(criteria, cb) {
-
+    
+    //make shallow copy of criteria so original does not get modified
+    if (Array.isArray(criteria) || _.isObject(criteria)) {
```
dmarcelino: I think we can simplify this to `if (_.isObject(criteria)) {` given that `_.isObject([]) === true`, actually I think there is no problem to remove the `if` and go with just `criteria = _.clone(criteria);`. I've seen that being done in other places.

Globegitter:
@devinivy @dmarcelino would it be possible to get this merged in? Makes it easier to carry on working with my `findAndModify` branch in my app.

devinivy:
It looks fine to me– I'm just a little surprised that cloning in `waterline/lib/waterline/query/finders/basic.js` didn't do you any good!  Maybe I'm misunderstanding how this whole thing is wired-up, but I thought that the criteria passed there went along this stack to the adapter method:
https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/basic.js#L63-L66
https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/operations.js#L69
https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/operations.js#L400
https://github.com/balderdashy/waterline/blob/master/lib/waterline/adapter/dql.js#L134

Globegitter:
@devinivy Unfortunately I also don't know how exactly it is wired up, but I added a `console.log` into that function and it wasn't displayed at all.

But maybe it is related because I am doing the `findOne` call in the same way as `findOrCreate` does here: https://github.com/balderdashy/waterline/blob/4653f8a18016d2bcde9a70c90dd63a7c69381935/lib/waterline/adapter/compoundQueries.js#L38 ?

devinivy:
Yup, that's why!  You're calling the adapter method directly, which is fine for your purposes (as long as you don't need any joins).  That's not the same exact findOne that you call from a collection.

dmarcelino:
I'm good with the change :white_check_mark:

Globegitter:
@devinivy you happy to merge as well?

PR #949, opened 2015-04-15
validate() presentOnly parameter can be used to specify attributes to validate.

marnusw:
The static `validate()` method on the model provides the `presentOnly` parameter to indicate that only the provided subset of values should be validated. When desiring to validate only a single modified value this can become problematic if that value depends on other attributes in the model, since the values of those attributes wouldn't be present at the time. Adding those values in is not ideal since it will cause them to be validated as well and requires extra code.

While this might not be a big issue on a server I am investigating the use of Waterline on the client in an isomorphic app. When some attributes require async validation to the server it becomes a big problem if I can't validate one property at a time, but then the issue in the first paragraph may come into play with the current implementation.

My solution is to allow passing the name of a single attribute to validate, or an array of attribute names, on the same presentOnly parameter. This is backwards compatible, and also allows for a wide range of use cases since the entire set of values can be provided with just the specified value(s) validated.

I have added a complete suite of unit tests for my changes.

### DIFF
--- a/lib/waterline/core/validations.js
+++ b/lib/waterline/core/validations.js
@@ -126,7 +126,9 @@ Validator.prototype.validate = function(values, presentOnly, cb) {
       } // Fall through to the default if the object is not an array
     default:
       // Any other truthy value.
-      validations = _.intersection(validations, Object.keys(values));
+      if (presentOnly) {
+        validations = _.intersection(validations, Object.keys(values));
+      }
   }
 
   function validate(validation, cb) {
--- a/lib/waterline/core/validations.js
+++ b/lib/waterline/core/validations.js
@@ -100,7 +100,8 @@ Validator.prototype.initialize = function(attrs, types, defaults) {
  * schema's validations using Anchor.
  *
  * @param {Object} values to check
- * @param {Boolean} presentOnly only validate present values
+ * @param {Boolean|String|String[]} presentOnly only validate present values (if `true`) or
+ * validate the specified attribute(s).
  * @param {Function} callback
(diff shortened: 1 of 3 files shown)


marnusw:
If you choose to merge this PR, perhaps you would consider another proposal towards client-side use of Waterline: _Selectively validating async validation rules or not._

In [`waterline\core\validations.js` lines 165-169](https://github.com/balderdashy/waterline/blob/b42bb4467058c5da8870e7adf4427da7c4dda1c0/lib/waterline/core/validations.js#L165-L169) synchronous vs. asynchronous functions (rules) are already handled separately. The only thing left to do is allow passing a flag to actually run the async methods or not.

Where this becomes handy client side is when you might check, for example, that a username is less than 10 characters (done locally) and that it is unique (with a call to the server). The former rule could be checked on every keystroke for immediate feedback, whereas the latter will only be processed once the user stops typing.

I'd love to hear you comments/suggestions, or if you want me to take a bash at it I can do that as well.

dmarcelino:
The code change and tests look good.

@devinivy, @tjwebb, what do you think?

devinivy:
Looks good :heavy_check_mark:

marnusw:
@dmarcelino Is this waiting for the next minor release? Is there a date scheduled for such a release?

(Since this doesn't introduce breaking changes perhaps it can go in a patch update?)

dmarcelino:
Thanks @marnusw!

marnusw:
Thanks for merging. :smile:

PR #509, opened 2014-07-02
Add `.pick()` and `.omit()` methods

jasonsims:
This PR adds the ability to apply projections to waterline queries. It's an extension of the work accomplished in  #162 by adding flexibility for how selection arguments are supplied as well as exclusions via `.omit()`. With this change, the adapters will always receive a standardized selection criteria object which they can process as needed for their specific query language. Here's an example of this object:
```js
// User.find().pick('name', 'age').exec(cb);
{ where: null, pick: { name: 1, age: 1 } }

// User.find().omit('name', 'age').exec(cb);
{ where: null, omit: { name: 0, age: 0 } }
```

Usage examples:
```js
// Include only 
User.find().pick('name', 'age').exec(cb);
User.find().pick(['name', 'age']).exec(cb);
User.find().pick({name: 1, age: 1}).exec(cb);

// Exclusion
User.find().omit('name', 'age').exec(cb);
User.find().omit(['name', 'age']).exec(cb);
User.find().omit({name: 0, age: 0}).exec(cb);
```

fixes #73

PR #952, opened 2015-04-17
Add `.pick()` and `.omit()` methods

### LINKED ISSUE #73: .select()
Support for both sorts of usages:

Traditional `SELECT`:
````javascript
User.find().select('name', 'age').exec(cb);
User.find().select(['name', 'age']).exec(cb);
````

`Prune` is an exclude (inverse of select)
````javascript
User.find().prune('password', 'apiKey').exec(cb);
User.find().prune(['password', 'id']).exec(cb);
````

dmarcelino:
This PR is the same as #509, but targeting `0.11`, props to @jasonsims. Please check #509 for the full discussion.

Original message:

> This PR adds the ability to apply projections to waterline queries. It's an extension of the work accomplished in  #162 by adding flexibility for how selection arguments are supplied as well as exclusions via `.omit()`. With this change, the adapters will always receive a standardized selection criteria object which they can process as needed for their specific query language. Here's an example of this object:
> ```js
> // User.find().pick('name', 'age').exec(cb);
> { where: null, pick: { name: 1, age: 1 } }
> 
> // User.find().omit('name', 'age').exec(cb);
> { where: null, omit: { name: 0, age: 0 } }
> ```
> 
> Usage examples:
> ```js
> // Include only 
> User.find().pick('name', 'age').exec(cb);
> User.find().pick(['name', 'age']).exec(cb);
> User.find().pick({name: 1, age: 1}).exec(cb);
> 
> // Exclusion
> User.find().omit('name', 'age').exec(cb);
> User.find().omit(['name', 'age']).exec(cb);
> User.find().omit({name: 0, age: 0}).exec(cb);
> ```
> 
> fixes #73

cc: @clarkorz, @mikermcneil, @mdunisch, @particlebanana, @sskyy, @devinivy, @tjwebb

### DIFF
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -94,6 +94,23 @@ Deferred.prototype.pick = function() {
 };
 
 /**
+ * Add a `omit` clause to the criteria object.
+ *
+ * Used for specifying a subset of attributes to `exclude` from the result.
+ *
+ * @param {Array|Object} List of keys to exclude
+ * @return this
+ * @chainable
+ */
+Deferred.prototype.omit = function() {
+  var omitCriteria = this._criteria.omit || {};
+  _.extend(omitCriteria, utils.parseProjectionCriteria(0, arguments));
+
+  this._criteria.omit = omitCriteria;
+  return this;
+}
+
+/**
  * Add a `joins` clause to the criteria object.
  *
  * Used for populating associations.
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -74,7 +74,23 @@ Deferred.prototype.populateAll = function(criteria) {
     self.populate(association.alias, criteria);
   });
   return this;
+};
 
+/**
+ * Add a `pick` clause to the criteria object.
+ *
+ * Used for specifying a subset of attributes to `include` within the result.
+ *
+ * @param {Array|Object} List of keys to include
+ * @return this
+ * @chainable
+ */
+Deferred.prototype.pick = function() {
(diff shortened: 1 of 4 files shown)


dmarcelino:
#73 actually asks for a `.select()` method so we should probably follow @devinivy's suggestion on https://github.com/balderdashy/waterline/pull/509#issuecomment-86695959:
> I think `pick` and `select` should be aliases.

tjwebb:
re: http://stackoverflow.com/a/29769995/291180

@dmarcelino would you mind adding a quick sentence or two in the docs about how the `select` criteria works?

dmarcelino:
@tjwebb, I can't believe you didn't know about it... to be honest I don't remember how I found out it existed. Sure, I'll look into adding it.

tjwebb:
>  to be honest I don't remember how I found out it existed. Sure, I'll look into adding it.

Yea I sort of knew about it, but I've never used it and didn't really know the API. Is it unit tested?

dmarcelino:
I don't think it is, there are some tests that use it ([quick search](https://github.com/balderdashy/waterline/search?utf8=✓&q=select)) but no particular unit test to explicitly test it.

zfrenchee:
Why isn't this a feature yet?

weyert:
Any news on this?

randallmeeker:
closed w/o comment?

chacliff:
personally i think that this should be structured just like mongo structures it... if your going to use "select" or "pick"... not omit it should just be <field>: 0 || 1, where 0 is to omit, and 1 is to include. by default if no selection is made everything is included. by default if a field is marked with 1 ID is still included unless explicitly marked 0.

Issue #441, opened 2014-05-23
Model.save() method should not always populate associations


mphasize:
The model.save() instance method should accept a parameter to indicate whether the model associations should be populated after saving.

Following use case: Implementing a publish method on a post model

In PostController I added a `publish` action loosely based off the `update` blueprint.

	publish: function ( req, res ) {
		var pk = req.options.id || ( req.options.where && req.options.where.id ) || req.param( 'id' );

		var values = null;
                // the json-structure used here is closer to json-api specs with { "post" : { ... values ... }}
		if ( req.body && req.body[ "post" ] ) {
			values = req.body[ "post" ];
			// Omit any params w/ undefined values
			values = _.omit( values, function ( p ) {
				if ( _.isUndefined( p ) ) return true;
			} );
		}
		if ( values === null ) {
			return res.serverError( "Please provide values for 'post' to be updated in the JSON API format, e.g. { post: { ... values ... } }" );
		}

		Post.findOne( pk ).exec( function found( err, post ) {
				if ( err ) return res.serverError( err );
				if ( !post ) return res.notFound();

				post.publish( values, function ( err, publishedPost ) {
					if ( err ) return res.negotiate( err );

					// If we have the pubsub hook, use the Model's publish method
					// to notify all subscribers about the update.
					if ( req._sails.hooks.pubsub ) {
						if ( req.isSocket ) {
							Post.subscribe( req, [ publishedPost ] );
						}
						Post.publishUpdate( pk, _.cloneDeep( publishedPost.toJSON() ), !req.options.mirror && req, {
							previous: post.toJSON()
						} );
					}

					var jsonAPI = {};
					jsonAPI[ "post" ] = publishedPost;
					res.ok( jsonAPI );

				} );
			}

		);

After **Post.findOne()** the returned `post` has flat (i.e. not populated) associations with only respective IDs returned.

After **post.save()** (called in post.publish()) the associations are populated in `publishedPost`.

This leads to an error in **Post.publishUpdate** with `previous` set, because the model state saved in previous doesn't have populated associations and the current state has.

The error is generated in https://github.com/balderdashy/sails/blob/master/lib/hooks/pubsub/index.js#L977

    if ( invalidId || invalidAlias || invalidAddedId ) {
      return sails.log.error(
        'Invalid usage of ' + this.identity +
        '`publishAdd(id, alias, idAdded, [socketToOmit])`'
    );

[...]

PR #954, opened 2015-04-18
Added ability to specify what to populate on save()

nkibler7:
I noticed that every time I called `.save()` on my models with large association lists, I was waiting seconds for it to finish. In fact, as the association list grew (i.e. posts for a user in a social network setting), performance degraded rapidly. This small feature addition allows an optional `options` parameter to be passed to `.save()`, so the new function header is: 

    model.save(cb, options);

This will allow for more options to be passed to the function in the future, if necessary. Specific to this pull request is the addition of the `populate` key.

The value can be an array of objects that contain details about the association you want to populate:

    // This will only populate the given associations 
    var pOpts = [
        {
            key: 'followers',
            query: {
                name: 'Mike',
                limit: 5
            }
        },
        {
            key: 'stream'
        }
    ];

    model.save(cb, { populate: pOpts });

Or the value can be a boolean, where `true` will populate all associations and `false` will populate none:

    // This will not populate any associations
    var pOpts = false;
    model.save(cb, { populate: pOpts });

The default value is `true` so existing code will not be affected by this change. If someone gets to writing the tests before I do, that would be fantastic (I am not familiar with this testing platform and it may take me some time to get to it). Let me know if anything should be changed!

Thanks,
Nate Kibler

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -33,8 +33,8 @@ module.exports = function(context, proto, cb, options) {
     deferred = defer();
   }
 
-  options = options || optDefaults;
   cb = cb || noop;
+  options = _.defaults(options, optDefaults) || optDefaults;
 
   /**
    * TO-DO:
@@ -190,12 +190,10 @@ module.exports = function(context, proto, cb, options) {
 
     // Build up a new query and re-populate specified associations
     var query = context.findOne(criteria);
-    var populate = options.hasOwnProperty('populate') ?
-      options.populate : optDefaults.populate;
 
-    if (populate) {
-      if (populate.constructor === Array) {
-        populate.forEach(function(pop) {
+    if (options.populate) {
+      if (options.populate.constructor === Array) {
+        options.populate.forEach(function(pop) {
           query.populate(pop.key, pop.query);
         });
       }
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -7,6 +7,9 @@ var removeAssociation = require('../associationMethods/remove');
 var hop = require('../../../utils/helpers').object.hasOwnProperty;
 var defer = require('../../../utils/defer');
 var noop = function() {};
+var optDefaults = {
+  populate: true
+};
 
 /**
  * Model.save()
@@ -29,6 +32,7 @@ module.exports = function(context, proto, cb) {
     deferred = defer();
   }
 
+  options = options || optDefaults;
   cb = cb || noop;
 
   /**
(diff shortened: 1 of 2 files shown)


tjwebb:
Hey @nkibler7 nice work! Thanks for the PR, we'll review this.

lib/waterline/model/lib/defaultMethods/save.js
```diff
@@ -29,6 +33,7 @@ module.exports = function(context, proto, cb) {
     deferred = defer();
   }
 
+  options = options || optDefaults;
```
tjwebb: I think `_.defaults(options, optDefaults)` would be more accurate here

nkibler7: Good call! I updated the code to use this. Admittedly I have never used lodash, so thanks for the heads up!

dmarcelino:
Thanks @nkibler7, my first impressions:
1. I would prefer the function signature `.save([options,] cb)` since having the callback as the last parameter is more conventional;
2. Considering what you wrote in your first paragraph it seems there is a serious performance issue with `.save()` and it would be good find and fix the root cause.

devinivy:
I imagine that the performance issues with `Model#save` are really issues with `Collection#update`.  Also, yes, we should change the function signature to be more conventional and consistent with the rest of waterline.

nkibler7:
Thanks for your comments everyone! I agree, a deep performance evaluation needs to be done on the `.save()` function as a whole. There are several areas that could be improved. My change alone (using my particular data setup) improved speeds from 5-6 seconds to ~400-500 ms. Though, speaking to the complexity of the actual operation performed, this could (and should) experience another 10x speed-up. I'll continue to evaluate the slow areas of this function in my spare time and I would greatly appreciate any help offered from anyone!

Quick question: how should I handle the optional parameter being first? My understanding is if the programmer didn't specify an `options` parameter, their callback function would get bound to the options parameter. I can think of many solutions, but I want to use the standard way practiced in the rest of this code base and I'm having trouble finding an example. Thanks for your help!

dmarcelino:
```javascript
function(options, cb) {
  if(!cb){
    cb = options;
    options = undefined;
  }
  //...
}
```

Should do it, here's another approach from the core code: https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/update.js#L29-L32

nkibler7:
Thanks! That's what I figured but I just wanted to make sure I follow the convention. Should this entire pull request be delayed until further investigation of the inefficiencies of the `.save()` and/or `.update()` functions, or should this be refined and pulled in as an initiator to that effort?

dmarcelino:
I've read #441 and the case being made for this and I think these changes are quite relevant. In a sense waterline API is a bit contradictory `.update()` does not populate but `.save()` does! :open_mouth: I'm convinced this is a good addition and we'll also need documentation.

@tjwebb, @devinivy, @particlebanana what are your thoughts?

devinivy:
This is a simple, powerful, backwards-compatible change– looks good to me!  I still do want to see the function signature change before merging too, though.  We need to look into performance issues, but this doesn't seem dependent on that.

dmarcelino:
> I still do want to see the function signature change before merging too, though.

@nkibler7, can you take care of that? Thanks.

nkibler7:
I would love to, but I'm actually in the middle of a huge move across the country. Between that and vacation, I won't have access to my dev PC for over a month. I can try to find some time in my free time to fix it, though.

dmarcelino:
No worries @nkibler7, enjoy your vacation! :wink:

tobalsgithub:
@nkibler7, don't want to steal your thunder here, but could really use this fix. Anyone have an issue if I take the changes here, do the signature change and create a separate pull request? Or is there an easier option?

dmarcelino:
> Or is there an easier option?

You could do a PR against @nkibler7's branch but given it may take a while for him to merge it I don't think there would be any issue with branching his changes and creating a new PR.

tobalsgithub:
Created https://github.com/balderdashy/waterline/pull/1055

dmarcelino:
@nkibler7, @tjwebb, @devinivy, @particlebanana this PR continues in #1055 with the function signature change.

PR #962, opened 2015-04-22
#887: change default migration to 'safe'

### LINKED ISSUE #887: migrate: 'alter' when using Waterline standalone

dmarcelino:
Fixes #887.

Depends on balderdashy/waterline-adapter-tests#50 or the integration tests will be broken.

### DIFF
--- a/lib/waterline/collection/defaults.js
+++ b/lib/waterline/collection/defaults.js
@@ -5,6 +5,6 @@
  */
 module.exports = {
 
-	migrate: 'alter'
+	migrate: 'safe'
 
 };
--- a/test/integration/Collection.ddl.js
+++ b/test/integration/Collection.ddl.js
@@ -14,6 +14,7 @@ describe('calling describe', function() {
 		bootstrapCollection({
 			adapter: Adapter,
 			properties: {
+			  migrate: 'alter',
 				attributes: {
 					name: 'string',
 					age: 'integer'


devinivy:
:heavy_check_mark: looks good!

tjwebb:
The tests are already broken in balderdashy/waterline-adapter-tests#50. Should I merge that anyway?

dmarcelino:
Looks like you've merged already... @tjwebb, the waterline-adapter-tests always fail because of issues in sails-postgresql (balderdashy/waterline-sequel#43), sails-mongo (balderdashy/sails-mongo#260), sails-mysql (https://github.com/balderdashy/waterline/issues/887#issuecomment-95210555) and sails-redis (this one will take more than 1 issue to fix).

The reason for not merging before balderdashy/waterline-adapter-tests#50 is that this is going to break (additional) integration tests on schemaful adapters.

particlebanana:
Just for posterity we need to look into sails and see if it relies on the default being alter. We should set it in sails so that it doesn't matter what the default here is but I'm not sure that's being done.

Either way lets be safe in get this in 0.11. Thanks everyone!

Issue #317, opened 2014-03-13
JSON.stringify() a model skips some properties


pixelfreak:
I'm on Sails 0.10. When I do JSON.stringify(someModel), some properties, specifically properties associated with other model is not outputted. How do I get it enumerated?

Thanks!

andrewjmead:
+1 this is still an issue for me. I'm using `0.10.16`.

wallaceicy06:
+1 this is still an issue for me as well

anasyb:
This is still an issue. A simple (but very bad) workaround is to set the .toJSON property of the returned record to anything other than a function, for example:
```javascript
User.find().exec(function(err, users){
  console.log(JSON.stringify(users[0]))
})
//{ id:1, name:'Susan'} //no post IDs!

User.find().exec(function(err, users){
  console.log(users[0].posts);
})
//['1', '2', '3', add:[Function add], remove:[Function remove]]  //but actually here they are!!

User.find().exec(function(err, users){
  users[0].toJSON="I won't allow you to use sails toJSON, just use the native stringification";
  console.log(JSON.stringify(users[0]))
})
//{ id:1, name:'Susan', posts:[1,2,3]} //post IDs are back!

```
As indicated by @particlebanana this is a problem with toObject, in particular it has to do with this [line, blame it for all this mess!](https://github.com/balderdashy/waterline/blob/master/lib/waterline/model/lib/defaultMethods/toObject.js#L145). Fixing the if condition or commenting it in your local waterline solves the problem, which is obviously a much better workaround, that is until it get official fixed.

This is very related to the *user-case* behind https://github.com/balderdashy/waterline/issues/532, although it is a completely separate issue. cc @mphasize

tjwebb:
@particlebanana this was marked In Progress forever ago. is there any update or did it call through the cracks

devinivy:
Any word here? @anasyb do you think your fix is solid enough to submit a PR?

anasyb:
@devinivy I had it running for a while with that line commented out... no notice of anything going wrong. Before a PR I suppose we need an understanding of what that's line purpose in life is/was.

Can someone run the tests with that line commented and post here which ones fail (if any)?

devinivy:
I wonder if `showJoins` is possibly not being accounted for on that line.

[...]

Issue #891, opened 2015-03-18
Model#toJSON returns an object with a toJSON method


weblogixx:
Hi,

I am encountering an recursion error when using sails with socket.io redis while updating models. This seems to be related to msgpack sizeof, which tries to call the toJSON method of my model entities. It also does not make a difference if I use a custom or the builtin toJSON method.

Software versions:
OS: Windows 7 (64 bit)
node.js -> v0.12.0
sails: 0.11
socket.io-redis: 0.1.4
sails-mysql: 0.10.11

This is the output I get when trying to update via socket.io:
```
backend\node_modules\sails-mysql\node_modules\mysql\lib\protocol\Parser.js:82
        throw err;
              ^
RangeError: Maximum call stack size exceeded
    at Date.valueOf (native)
    at baseClone (backend\node_modules\sails\node_modules\waterline\node_modules\lodash\dist\lodash.js:765:29)
    at Function.cloneDeep (backend\node_modules\sails\node_modules\waterline\node_modules\lodash\dist\lodash.js:1799:14)
    at backend\node_modules\sails\node_modules\waterline\lib\waterline\model\lib\defaultMethods\toObject.js:114:26
    at Array.forEach (native)
    at toObject.addProperties (backend\node_modules\sails\node_modules\waterline\lib\waterline\model\lib\defaultMethods\toObject.js:112:27)
    at new module.exports (backend\node_modules\sails\node_modules\waterline\lib\waterline\model\lib\defaultMethods\toObject.js:34:8)
    at prototypeFns.toObject (backend\node_modules\sails\node_modules\waterline\lib\waterline\model\index.js:30:14)
    at prototypeFns.toJSON (backend\node_modules\sails\node_modules\waterline\lib\waterline\model\index.js:112:20)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:27)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:14)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:14)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:14)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:14)
    at sizeof (backend\node_modules\socket.io-redis\node_modules\msgpack-js\msgpack.js:466:14)
[...]

PR #969, opened 2015-04-23
modify toObject so that you get a POJO back

particlebanana:
this *should* help with the recursion issues and solve #891

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/toObject.js
+++ b/lib/waterline/model/lib/defaultMethods/toObject.js
@@ -22,21 +22,42 @@ var hasOwnProperty = utils.object.hasOwnProperty;
 
 var toObject = module.exports = function(context, proto) {
 
+  var self = this;
+
   this.context = context;
   this.proto = proto;
 
   // Hold joins used in the query
   this.usedJoins = [];
 
-  this.object = Object.create(proto.__proto__);
+  // Create an object that can hold the values to be returned
+  this.object = {};
 
+  // Run methods to add and modify values to the above object
   this.addAssociations();
   this.addProperties();
   this.makeObject();
   this.filterJoins();
   this.filterFunctions();
 
-  return this.object;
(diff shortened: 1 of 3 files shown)


particlebanana:
call stack issues here https://github.com/balderdashy/sails/issues/2813 and https://github.com/balderdashy/sails/pull/2647

devinivy:
Will look over this asap, but at a very cursory glance this seems okay to me.  Someone said that this could affect blueprints with the socket hook in sails (https://github.com/balderdashy/waterline/issues/891#issue-62683601).  I'd rather that issue not hold this bug up, but I'm just wondering if it's being addressed.

particlebanana:
I have that PR out in Sails which is a workaround in the hook but this should fix that up unless I misunderstood the sails issue.

dmarcelino:
This will probably also fix #317.

@particlebanana, can't see anything wrong with this :white_check_mark:

devinivy:
From the tests it looks like associations are maintained, which was my only concern.  Seems okay!

dmarcelino:
I only have one minor concern... in the below line:
``` javascript
var _obj = JSON.parse(JSON.stringify(this.object));
```
an error will be thrown if the object has any circular references. While circular references are not allowed in JSON we will be limiting `toObject()` with the same constraint. Not huge as many adapters don't return circular references and in, many cases, they are not desired because the next move is usually to jsonfy the object.

particlebanana:
Ok thanks guys, gonna pull this in and do some real world testing. @dmarcelino as for the circular references, technically you could handle that without ever calling toObject - or more likely before. The default model behavior is to call this before stringifying it and sending to the client so it should help clear up a lot of issues. Let's cross the circular ref issue when/if it comes up.

dmarcelino:
> Let's cross the circular ref issue when/if it comes up.
Agreed!

Issue #698, opened 2014-10-23
Waterline 0.10.11 passing array as string to afterUpdate


chadxz:
When I update a record in the database (using 'sails-mysql' adapter), the `values` passed to the afterUpdate lifecycle callback looks like this:

```
{ id: '4d094885-7e6f-4454-91f0-6a55e3d1b6eb',
  accountId: 'webhooksControllerTestAccountId',
  appId: 'webhooksControllerTestAppId',
  endpointId: 'updatedEndpointId',
  webhookURL: 'updatedWebhookURL',
  groups: '["updatedGroups"]',
  createdAt: Thu Oct 23 2014 15:34:12 GMT-0500 (CDT),
  updatedAt: Thu Oct 23 2014 15:34:12 GMT-0500 (CDT) }
```

As you can tell from the model definition below, the groups field is an array. I have checked to ensure that i'm passing it an array when its going in, and reading it back out of the database the value is an array, but the value passed to the afterUpdate is a string

I can probably fix this and issue a PR, but i was wondering where I should start looking for the issue... would the code responsible for serializing arrays to string and back be in waterline or sails-mysql?

model definition:
```
module.exports = {

    attributes: {
        id: {
            type: 'string',
            primaryKey: true
        },
        accountId: {
            type: 'string',
            required: true
        },
        appId: {
            type: 'string',
            required: true
        },
        endpointId: {
            type: 'string',
            required: true
        },
        webhookURL: {
            type: 'string',
            required: true
        },
        groups: {
            type: 'array',
            defaultsTo: []
        },
        toJSON: function () {
            var obj = this.toObject();
            delete obj.createdAt;
            delete obj.updatedAt;
            return obj;
        }
    }
```

devinivy:
As I understand it, JSON/Array data types should be passed to the afterUpdated lifecycle as an object/array respectively, rather than a string.  That sounds reasonable.  @chadxz mind changing the title to reflect this?  Anybody mind writing a failing test [here](https://github.com/balderdashy/waterline/tree/master/test/unit/callbacks)?

chadxz:
@devinivy on it

chadxz:
@devinivy this doesn't seem to be an issue with the waterline layer, but of the sails-mysql adapter. Waterline is just passing the values to afterUpdate what it got from the adapter.

chadxz:
i'm gonna keep trying for a bit

chadxz:
nevermind, i'm seeing the issue now

[...]

PR #971, opened 2015-04-24
afterUpdate callback not provided the correct schema

chadxz:
Here is a failing test for #698

### DIFF
--- a/test/unit/callbacks/afterUpdate.update.js
+++ b/test/unit/callbacks/afterUpdate.update.js
@@ -1,0 +1,78 @@
+var Waterline = require('../../../lib/waterline'),
+  assert = require('assert');
+  _ = require('lodash');
+
+describe('.afterUpdate()', function() {
+  var User, afterUpdateValues = null;
+
+  before(function (done) {
+    var waterline = new Waterline();
+    var Model = Waterline.Collection.extend({
+      identity: 'user',
+      connection: 'foo',
+      attributes: {
+        name: 'string',
+        employed: 'boolean',
+        meta: 'json',
+        hobbies: 'array'
+      },
+
+      afterUpdate: function (values, cb) {
+        afterUpdateValues = _.cloneDeep(values);
+        cb();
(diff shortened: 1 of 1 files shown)


test/unit/callbacks/afterUpdate.update.js
```diff
@@ -0,0 +1,78 @@
...
+
+    // Fixture Adapter Def
+    var adapterDef = {
+      update: function (con, col, criteria, values, cb) {
+        // emulate values coming back from (for example) sails-mysql adapter
+        Object.keys(values).forEach(function (key) {
+          var val = values[key];
+          if (typeof val === 'object') {
+            // will stringify arrays and objects
+            values[key] = JSON.stringify(val);
```
dmarcelino: This forces the values to change type, so I think this test will always fail.

test/unit/callbacks/afterUpdate.update.js
```diff
@@ -0,0 +1,78 @@
...
+        cb();
+      }
+    });
+
+    waterline.loadCollection(Model);
+
+    // Fixture Adapter Def
+    var adapterDef = {
+      update: function (con, col, criteria, values, cb) {
+        // emulate values coming back from (for example) sails-mysql adapter
```
dmarcelino: I think this is MySQL specific, so this test should be in the waterline-adapter-tests.

dmarcelino:
Hi @chadxz, after looking at this I don't think this is the right place/way to test this issue as in your test it forces invalid values to be passed to `afterUpdate`. In other words, this test would promote the wrong fix.

Please take a look at balderdashy/waterline-adapter-tests#73, I believe that's the correct way of testing this issue. The [build results](https://travis-ci.org/dmarcelino/waterline-adapter-tests/jobs/62814934#L550) show this issue only happens on sails-mysql and that's where this needs to be fixed.

I'm closing this for now, please re-open if you disagree of if I missed something.

Thanks

PR #972, opened 2015-04-24
860: return error on findOne(undefined).populate()

### LINKED ISSUE #860: Model.findOne(undefined).populate('some-1-to-1-mode') returns a record
Model.findOne(undefined) throws Error: .findOne() requires a criteria.

But Model.findOne(undefined).populate('some-1-to-1-mode') returns a record (in my case, the record with id = 1)

The current solution is to always use Model.findOne({id: undefined})

dmarcelino:
Fixes #860.

`findOne(undefined)` returns an error but `findOne(undefined).populate()` does not. This PR makes both uses of `.findOne()` consistent by always throwing error.

### DIFF
--- a/lib/waterline/query/finders/basic.js
+++ b/lib/waterline/query/finders/basic.js
@@ -55,8 +55,10 @@ module.exports = {
 
     // If there was something defined in the criteria that would return no results, don't even
     // run the query and just return an empty result set.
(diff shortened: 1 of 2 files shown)


devinivy:
Looks okay!

dmarcelino:
Thanks again @devinivy!

PR #977, opened 2015-04-26
Extend types through the adapter

### LINKED ISSUE #819: are there anyways to extend types in our own adapters?

dmarcelino:
Fixes #819 and is an alternative to PR #832.

In similar fashion to Custom Types (PR #72) used in the model definition, this PR introduces the option to extend waterline's types in the adapter following the same format as Custom Types / [anchor rules](https://github.com/balderdashy/anchor/blob/master/lib/match/rules.js). For an adapter to make use of this it only requires to have a field named `types` in its adapter.js, example:
``` javascript
  types: { 
    myDbSpecialNumberFormat: function(val){ return !isNaN(val); } 
  }
```

Now, in the model definition the user would be able to use:
```javascript
attributes: {
  spaceDistance: {
    type: "float",
    myDbSpecialNumberFormat: true
  }
}
```

Compared to #832 this change has the advantages of being more similar to Custom Types format and it includes validations which #832 does not.

cc: @shekhei, @devinivy, @GaryLCoxJr

### DIFF
--- a/lib/waterline/core/index.js
+++ b/lib/waterline/core/index.js
@@ -84,11 +84,13 @@ Core._initialize = function(options) {
   }
 
   var reservedAttributes = adapterInfo.reservedAttributes || {};
+  var defaultValidations = this.defaults.validations || {};
+  var adapterTypes = adapterInfo.types || {};
 
   // Initialize internal objects from attributes
   this._schema.initialize(this._attributes, this.hasSchema, reservedAttributes);
   this._cast.initialize(this._schema.schema);
-  this._validator.initialize(_validations, this.types, this.defaults.validations);
+  this._validator.initialize(_validations, this.types, adapterTypes, defaultValidations);
 
   // Set the collection's primaryKey attribute
   Object.keys(schemaAttributes).forEach(function(key) {
--- a/lib/waterline/core/validations.js
+++ b/lib/waterline/core/validations.js
@@ -53,8 +53,13 @@ var Validator = module.exports = function(adapter) {
  * }
  */
 
-Validator.prototype.initialize = function(attrs, types, defaults) {
+Validator.prototype.initialize = function(attrs, types, adapterTypes, defaults) {
   var self = this;
+
+  if(!defaults){
+    defaults = adapterTypes;
+    adapterTypes = undefined;
+  }
 
   defaults = defaults || {};
 
@@ -66,6 +71,10 @@ Validator.prototype.initialize = function(attrs, types, defaults) {
   if(defaults.ignoreProperties && Array.isArray(defaults.ignoreProperties)) {
     this.reservedProperties = this.reservedProperties.concat(defaults.ignoreProperties);
   }
+  
+  // add adapter type definitions to anchor
+  adapterTypes = adapterTypes || {};
+  anchor.define(adapterTypes);
 
   // add custom type definitions to anchor
   types = types || {};
--- a/test/integration/Collection.validations.js
+++ b/test/integration/Collection.validations.js
@@ -42,6 +42,10 @@ describe('Waterline Collection', function() {
 
           password: {
             type: 'password'
+          },
+          
+          age: {
+            type: 'age'
           }
         }
       });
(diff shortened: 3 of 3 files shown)


devinivy:
A few questions!

- Can you give an example of how this would this be used in a model definition?
- How are we dealing with fallbacks to core waterline types and specifying those fallbacks in a model definition?
- How is this different than using a core waterline type with custom validation?

dmarcelino:
* Assuming you've defined a custom type in your adapter named "myDbSpecialNumberFormat" you could do the following in a model definition:
  
  ```javascript
  {
    attributes: {
      spaceDistance: {
        type: "myDbSpecialNumberFormat"
      }
    }
  }
  ```
  Another example in the test I've added: https://github.com/dmarcelino/waterline/blob/adapter_types/test/integration/Collection.validations.js#L47-L49
* What you mean by fallbacks? The case where the adapter overwrites a waterline core type? It should work in similar manner as Custom Types defined in the model.
* It's pretty similar, the difference would be that this gives adapter developers freedom to create their own types that are not shared across the waterline universe of types.

devinivy:
I think that there still needs to be some fallback– I.e. (and this is what I mean by "fallback") what happens when I want to switch adapters and the type specified in my model is not supported?  I have some concerns about model portability there.  I'd rather see adapters providing their own validations to hook into, or something like that.  The waterline types are nearly analogous to types in javascript, because everything that comes out of the database will be represented as one of those types in your application!  On top of that, validations make sure the data fits nicely into the database.  Curious to hear @particlebanana's thoughts on this.

dmarcelino:
@devinivy, I see what you mean now. I see this feature as adapters' Custom Methods or `ignoreProperties` (balderdashy/waterline-docs#66) which are defined per adapter and extend waterline functionality in a way which is not necessarily compatible with other adapters. A user should still able to do:
```javascript
attributes: {
  spaceDistance: {
    type: "float",
    myDbSpecialNumberFormat: true
  }
}
```
which would provide a fallback of sorts. The only problem with this is it would make it harder for the adapter to figure out it should use `type: "myDbSpecialNumberFormat"` instead of `type: "float"`. I can add a test for that if you are interested.

PS: I would also like to hear @tjwebb and @shekhei thoughts.

devinivy:
What if validations had a property named `type`.
```node
var geoSpatialValidator = function(obj) {

   return isOnEarth(obj);
}

geoSpatialValidator.type = 'geo';
```

dmarcelino:
I'm not sure if/how the adapter receives that info... From what I vaguely remember the adapter gets a schema which has the types after being transformed by waterline core (example: e-mail type is converted to string) and the actual model definition the user wrote. In order to something like that waterline core would have to process it and handle it, right?

devinivy:
The adapter could find the validator with a little bit of effort, but there's no good way to do this currently.  My suggestion would require some sort of PR to expose this validation/type info to the adapter.  Or better yet, also expose validations from the adapter to be used by the model. I imagine some convention, like `require('sails-mysql/types')` which would contain a suite of mysql types in the form of validation functions.

dmarcelino:
I won't mind pushing for:
```javascript
attributes: {
  spaceDistance: {
    type: "float",
    myDbSpecialNumberFormat: true
  }
}
```
This is feasible and even though it would require the adapter's developer a little bit more effort it would provide a fallback. The advantage of this is that all the necessary pieces are already in place. In the adapter instead of one looking just at the `schema` the model definition would also have to be considered.

dmarcelino:
@devinivy, another more complex but more explicit option is to create the concept of `dbType` or `customDbType` as such:
```javascript
attributes: {
  spaceDistance: {
    type: "float",
    dbType: myDbSpecialNumberFormat
  }
}
```

Then we could pass `dbType` in the `schema` object (to make the adapter's developer life easier) and we  could apply the correct validation.

dmarcelino:
@devinivy, if you like my last comment's suggestion (https://github.com/balderdashy/waterline/pull/977#issuecomment-96792387) I'll work on a POC, shouldn't be to hard and I actually think it's the better approach.

devinivy:
I like it!  Another option, API-wise, is to accept an object in `type`, and it could have separate fields for typing options, i.e. typing options per adapter or a list of matching types.  Could be a good way to keep type options in one place.  If you add a new field `dbType`, it will require a change to waterline-schema.  All that said, I think a new field would be okay too, esp for a POC.

dmarcelino:
@devinivy, check #981 please.

Replacing `type` for an object would be a difficult feat to achieve API-wise without introducing breaking changes for adapters (I think).

devinivy:
Will check!  And good point!

dmarcelino:
Closing this for now as PR #977 seems to be the best way to go ahead.

PR #980, opened 2015-04-27
createEach: use async.map instead of async.eachSeries to improve performance

dmarcelino:
In #795, the `async.each` was replaced by `async.eachSeries` to ensure order of creation and make tests pass (for community adapters?), this however introduces a performance penalty which in my opinion and others is not worth:

@tjwebb in https://github.com/balderdashy/sails-postgresql/issues/128#issuecomment-68575680
> `.eachSeries` will slow things down, though, since each create will then be done sequentially. Should the order really matter, here?

@kevinburkeshyp in https://github.com/balderdashy/sails-postgresql/issues/128#issuecomment-95970705:
> would `async.map` help? 

@devinivy in https://github.com/balderdashy/waterline-adapter-tests/pull/27#issuecomment-96682857:
> I also would like to see `Async#map` being used as @dmarcelino mentioned.

### Pros and cons
Using [`async.map`](https://github.com/caolan/async#maparr-iterator-callback) instead of `async.eachSeries` will ensure `.create([])` returns its results in order (issues balderdashy/sails-postgresql#128 and balderdashy/waterline-adapter-tests#27) without limiting performance. For situations where creation order is sensitive users can still chain independent `create()` commands in an `async.eachSeries` calls.

### Tests
Lastly, regarding breaking tests, me and @devinivy have been fixing them in the waterline-adapter-tests (examples: balderdashy/waterline-adapter-tests#43, balderdashy/waterline-adapter-tests#52). I think it's best to fix the tests than to restrain waterline performance.

cc: @imevs, @particlebanana, @devinivy, @tjwebb, @brandonsimpson, @kevinburkeshyp, @atiertant

### DIFF
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -65,7 +65,7 @@ module.exports = {
       });
     }
 
-    async.eachSeries(valuesList, create, function(err) {
+    async.map(valuesList, create, function(err) {
       if(err) return cb(err);
       cb(null, records);
     });


devinivy:
I think this is the right idea.  The only concern would be exhaustion of db connections, but the adapters should deal with queuing on their own!

dmarcelino:
Pulling @particlebanana's feedback to this thread (https://github.com/balderdashy/waterline/pull/795#issuecomment-96720395):
> Yeah I'm down. Ideally we check if the adapter defines their own `createEach` method and use that and just use the `.map` as a fallback. This way we could get this to a single insert query in pg and mysql.

dmarcelino:
@imevs, @brandonsimpson, any thoughts regarding this?

kevinburkeshyp:
I appreciate the work done so far to address this, and this may be a crazy idea, but Postgres at least has support for a batch insert... (see "Insert multiple rows" here) http://www.postgresql.org/docs/9.3/static/sql-insert.html

particlebanana:
@kevinburkeshyp I mentioned that [here](https://github.com/balderdashy/waterline/pull/795#issuecomment-96720395). We should check if the adapter implements it's own `createEach` method and if so let it do it's thing. Otherwise fallback to the wl `createEach` method.

brandonsimpson:
@dmarcelino I'm ok with whatever solution is available to insert records in order. I'm not sure of an instance where I would be inserting so many records that it would be a performance issue. Also, it's not the result order that was my issue, my issue was the actual records being inserted out of order in the database. Sometimes that's important, especially if keys are generated by the database and are used elsewhere and depend on the order of data. Some people may prefer a slight performance boost over organized data but I prefer having my data organized how I anticipate it being created... which comes from a relational db background I suppose.

devinivy:
@brandonsimpson is this useful for programmers, computers, neither, or both?  I would personally never assume the order that records are inserted into the database.  I think using `map` merely adds a nice property of the method whereby the order of the returned results is predictable– it's a sort of luxury.  An adapter could always implement `createEach` if it's a special kind of database where the order of batch inserts has significance.

dmarcelino:
I think not limiting performance should be the default for a few reasons like waterline is async and some datastores have string ids.

Anyone wanting to preserve insertion order can do something like:
``` javascript
var values = [{...}, {...}];
async.eachSeries(values, function(val, next){
  model.create(val, next);
}, cb);
```

brandonsimpson:
@devinivy Just from my perspective, I think it's more useful for programmers and data analysts. Obviously in a program you can reference a record however you want from wherever you want, and I don't know how others are using waterline. But for example, as a human looking at data reports and building e-commerce systems, inserting order details in a mysql database and using the primary key as the sort order of the order details in future queries is important without adding an additional step to create some type of sequential identifier on the records when the database can do it perfectly fine on it's own. The details of the order are important to keep the same from cart to storage to reporting. 

If waterline isn't capable of doing this it's ok... I was already doing things with bluebird to get the order inserted correctly as a workaround. The last update to creatEach seemed to work fine for me, and I didn't see any performance issues personally but didn't dig in too much yet. If nothing else it would be nice maybe even as an optional flag on .createEach() or .create() if the inserts should be sequential.

dmarcelino:
I would see no issue in `.create()` being extended to have an additional `options` parameter  to allow that behaviour such as:
``` javascript
.create( criteria, [options, callback] )
```

What do the others think?

brandonsimpson:
@dmarcelino I think if `.create()` already accepts arrays and can do multiple inserts then `.createEach()` is redundant if it's not doing something more such as create each record sequentially. I think adding an `options` param to `.create()` would ultimately allow you to drop `.createEach()` completely.

dmarcelino:
@brandonsimpson, you are right about that. `.findOrCreateEach()` was also deprecated in favour of `findOrCreate([])` so I guess `.createEach()` should follow the same faith. There is already a discussion about which helper methods to drop in `0.11`: #931. I'll add your comment there.

dmarcelino:
> I mentioned that [here](https://github.com/balderdashy/waterline/pull/795#issuecomment-96720395). We should check if the adapter implements it's own `createEach` method and if so let it do it's thing. Otherwise fallback to the wl `createEach` method.

@particlebanana, @kevinburkeshyp, waterline is already calling `adapter.createEach()` if the adapter implements it (if I read [aggregateQueries.js](https://github.com/balderdashy/waterline/blob/master/lib/waterline/adapter/aggregateQueries.js#L29-L31) correctly) and sails-postgresql does implement `createEach()`: [adapter.js](https://github.com/balderdashy/sails-postgresql/blob/master/lib/adapter.js#L402).

dmarcelino:
Merged this and now sails-postgresql and sails-mysql break the **.create() should return rows in the correct order when creating multiple rows** test :confused:
https://travis-ci.org/balderdashy/waterline-adapter-tests/jobs/60719805#L481

dmarcelino:
Guys, please take a look at PR #986 which completes this fix.

PR #981, opened 2015-04-28
Extend types through the adapter using dbType

### LINKED ISSUE #819: are there anyways to extend types in our own adapters?

dmarcelino:
Fixes #819 and follows on the discussion in PR #977. This is also an alternative to PR #832. Do note this breaks the tests because it requires anchor changes, details below.

In similar fashion to Custom Types (PR #72) used in the model definition, this PR introduces the option to extend waterline's types in the adapter following the same format as Custom Types / [anchor rules](https://github.com/balderdashy/anchor/blob/master/lib/match/rules.js). For an adapter to make use of this it only requires to have a field named `types` in its adapter.js, example:
``` javascript
// adapter.js
// ...
  types: { 
    mydbspecialnumberformat: function(val){ return !isNaN(val); } 
  }
  // in similar fashion to `type` the key must be lower case
```

Now, in the model definition the user would be able to use:
```javascript
// some model definition
// ...
attributes: {
  spaceDistance: {
    type: "float",
    dbType: "myDbSpecialNumberFormat"
  }
}
```

### Comparison to previous PRs

Compared to PR #977 this change has the advantage of being transparent for adapters not providing validations for a given `dbType` while allowing that adapter to use the `type` as a fallback.

Compared to #832 this change has the advantages of being more similar to Custom Types format and it includes validations which #832 does not.

### Tasks

Please note, this PR depends on balderdashy/anchor#77 being merged and published. Hence the tasks to merge this are:
- [x] Review (and merge) balderdashy/anchor#77;
- [x] Publish anchor to npm;
- [x] Review (and merge) balderdashy/waterline-schema#26
- [x] Publish waterline-schema to npm;
- [ ] Re-run this PR tests and make sure they pass;
- [ ] Update waterline package.json to use the latest versions of anchor and waterline-schema;
- [ ] Merge this.

cc: @shekhei, @devinivy, @GaryLCoxJr, @tjwebb, @particlebanana

### DIFF
--- a/lib/waterline/core/index.js
+++ b/lib/waterline/core/index.js
@@ -84,11 +84,13 @@ Core._initialize = function(options) {
   }
 
   var reservedAttributes = adapterInfo.reservedAttributes || {};
+  var defaultValidations = this.defaults.validations || {};
+  var adapterTypes = adapterInfo.types || {};
 
   // Initialize internal objects from attributes
   this._schema.initialize(this._attributes, this.hasSchema, reservedAttributes);
   this._cast.initialize(this._schema.schema);
-  this._validator.initialize(_validations, this.types, this.defaults.validations);
+  this._validator.initialize(_validations, this.types, adapterTypes, defaultValidations);
 
   // Set the collection's primaryKey attribute
   Object.keys(schemaAttributes).forEach(function(key) {
--- a/lib/waterline/core/schema.js
+++ b/lib/waterline/core/schema.js
@@ -90,6 +90,12 @@ Schema.prototype.objectAttribute = function(attrName, value) {
         // Allow validation types in attributes and transform them to strings
         attr.type = ~types.indexOf(value[key]) ? value[key] : 'string';
         break;
+        
+      // Set schema[attribute].dbType
+      case 'dbType':
+        // Allow custom adapter validation types in attributes
+        attr.dbType = value[key];
+        break;
 
       // Set schema[attribute].defaultsTo
       case 'defaultsTo':
--- a/lib/waterline/core/validations.js
+++ b/lib/waterline/core/validations.js
@@ -53,8 +53,13 @@ var Validator = module.exports = function(adapter) {
  * }
  */
 
-Validator.prototype.initialize = function(attrs, types, defaults) {
+Validator.prototype.initialize = function(attrs, types, adapterTypes, defaults) {
   var self = this;
+
+  if(!defaults){
+    defaults = adapterTypes;
+    adapterTypes = undefined;
+  }
 
   defaults = defaults || {};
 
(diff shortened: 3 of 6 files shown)


devinivy:
Good lookin stuff, @dmarcelino !!  Does waterline-schema not need any changes?  I was imagining here: https://github.com/balderdashy/waterline-schema/blob/master/lib/waterline-schema/utils.js#L8

dmarcelino:
Thanks @devinivy, in my tests (even a crude one with sails-orientdb) I didn't need to change waterline-schema to see `dbType` on my adapter. Not sure were that list you mentioned is used because I don't see `size` there, for example... :confused:

dmarcelino:
PR balderdashy/waterline-schema#26 issued and list of task updated. Better safe than sorry.

devinivy:
Can we ramp this back up?  @particlebanana might need your input here.

tjwebb:
+1

weyert:
Would be great if this could get merged.

RWOverdijk:
:+1:

tjwebb:
fwiw, I merged this locally and have been using it in one of our projects (at Balderdash) and it seems to work really well. 

@particlebanana anchor and waterline-schema need to be published before the tests will pass on this PR

particlebanana:
So remind me how this works again. Is `dbType` essentially just a way to pass through types to the adapter for migrations? From the examples in the tests it looks like these are more runtime validations done at the adapter level.

@tjwebb what's the use case you used it on for the project and how did it work?

```javascript
types: { customdbtype: function(val){ return !isNaN(val) && val > 0; } }
```

tjwebb:
@dmarcelino this includes support for dbType foreign keys: https://github.com/waterlinejs/waterline/pull/1/files

@particlebanana I'm using it for two things currently:
1. I'm using the geometry type for spatial queries and spatial indexing in PostGIS
2. Makes it easier to use the native postgres `uuid` type, and using the postgres `generate_uuid_v4()` function to generate uuids, rather than in node

particlebanana:
This and things like [this](https://github.com/balderdashy/sails-postgresql/pull/202#issuecomment-149348924) that are only used for database auto migrations really make me wish we had a good migrations solution outside of Waterline.

devinivy:
@particlebanana it's a little beyond a migrations issue, maybe?  Since multiple column types in the actual DB might have the same representation as far as js/waterline is concerned, the `type` in waterline lossily represents the data format or structure.  On one hand that's fine– waterline core's job is to deal with un/serialization and validation of those simple types, but there also needs to be some way to relate to the adapter layer some extra information about the field types.  That could be for all sorts of reasons, including but not limited to building queries.  If the adapter is going to query the database, it needs to know what type of column it's querying.

particlebanana:
@devinivy not sure how common it is across adapters but I know in the postgresql adapter it calls `describe` for each table during `registerConnection` so it has an internal mapping of the db schema.

nkibler7:
Why isn't this merged yet?

RWOverdijk:
Because the checks failed, I think

iwasrobbed:
Any chance someone can champion this across the finish line? Perhaps @particlebanana or @mikermcneil

PR #982, opened 2015-04-28
867: `required: false` prevents all other validators from running fix

### LINKED ISSUE #867: `required: false` prevents all other validators from running, even when attribute has non-null value
When I set `required: false` in a model attribute's configuration, none of the other validations run, even when I am setting the attribute to a non-null, non-undefined value.

For example:
```
// here's your model config for User
module.exports = {
    //...
    attributes: {
        secondaryEmail: {
            required: false,
            email: true // this validation will not execute
        }
    }
    //...
}
 
// and you try to do this
User.update({id: userId}, {
    secondaryEmail: 'not an email' // should be rejected by "email" validator
}).exec(function(err, updated) {
    assert(err == null); // the update succeeded.  Is this intended behavior?
    //...
});
```

Here is the relevant waterline source code:
https://github.com/balderdashy/waterline/blob/master/lib/waterline/core/validations.js#L132-141

Is this intended behavior or a bug?

dmarcelino:
Fixes #867.

In the below case the e-mail validation does not run.

``` javascript
    attributes: {
        secondaryEmail: {
            required: false,
            email: true // this validation will not execute
        }
    }
```

The issue was traced to [validations.js#L138-L141](https://github.com/balderdashy/waterline/blob/master/lib/waterline/core/validations.js#L138-L141):
``` javascript
    // if required is set to 'false', don't enforce as required rule
    if (curValidation.hasOwnProperty('required')&&!curValidation.required) {
        return cb();
    }
```

This PR adds a failing test, 2 regressions tests and fixes this issue.

### DIFF
--- a/lib/waterline/core/validations.js
+++ b/lib/waterline/core/validations.js
@@ -135,11 +135,6 @@ Validator.prototype.validate = function(values, presentOnly, cb) {
       if(value === null || value === '') return cb();
     }
 
-    // if required is set to 'false', don't enforce as required rule
-    if (curValidation.hasOwnProperty('required')&&!curValidation.required) {
-        return cb();
-    }
-
     // If Boolean and required manually check
     if(curValidation.required && curValidation.type === 'boolean' && (typeof value !== 'undefined' && value !== null)) {
       if(value.toString() == 'true' || value.toString() == 'false') return cb();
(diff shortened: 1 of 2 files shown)


devinivy:
:heavy_check_mark:

Issue #925, opened 2015-04-02
Address integrator issue (Cross database associations)


devinivy:
A potential issue with the integrator was brought-up by @particlebanana in https://github.com/balderdashy/waterline/issues/664#issuecomment-57248273 , issue #664.  This is a technical issue that needs review and a potential fix.  If @particlebanana could chime-in what exactly to look for, that would be very useful!

**Pending tasks:**
- [x] 1. PR #984 merged
- [x] 2. PR #985: unit tests may need to be revised 
- [x] 3. PR balderdashy/sails-mysql#212 merged
- [x] 4. PR balderdashy/sails-postgresql#158 merged
- [ ] 5. balderdashy/waterline-adapter-tests#64
  - [x] Enable cross-adapter tests in sails-mongo balderdashy/sails-mongo#282
  - [x] Enable cross-adapter tests in sails-memory balderdashy/sails-memory#22
  - [x] Enable cross-adapter tests in sails-disk balderdashy/sails-disk#43
  - [x] Enable cross-adapter tests in sails-mysql balderdashy/sails-mysql#221
  - [x] Enable cross-adapter tests in sails-postgresql balderdashy/sails-postgresql#160
- [x] 6. Fix Populate m:m cross adapter/connection issue #1014

dmarcelino:
Could also fix:
* Association .populate() only populates defined attributes #366
* weird results in populated data #503
* Join failing across adapters #833

Possible solution, PR: Allow Associations to Use 'type' as a Name #789

devinivy:
@particlebanana do you have a moment to provide some insight as to general integrator issues to look into?

dmarcelino:
Hey @devinivy, @particlebanana, I have good news and bad news:
* The good news are I've managed to create a comprehensive test suite for associations across multiple connections (same adapter) by re-using the existing associations tests: https://github.com/balderdashy/waterline-adapter-tests/commit/dc7c871377b2a6e2c89f1719c9675309cf966c6e
* The bad news is that a lot of tests are breaking. Even for sails-memory which passes more tests it's breaking 26 and passes 40. Most breaking tests are nested operations: https://travis-ci.org/balderdashy/waterline-adapter-tests/jobs/60035080#L937

Let me know what you think.

dmarcelino:
PR #848, makes 8 additional cross db tests pass, can you guys take a look?

dmarcelino:
[...]

PR #984, opened 2015-05-01
Integrator kungfu

particlebanana:
Ok this should solve the issues that I currently know about in the integrator. 

RE: https://github.com/balderdashy/waterline/issues/925

* It fixes an issue when using column names on model associations where the transformer was getting the wrong data.

* It fixes an issue when using many-to-many associations across adapters where children weren't being populated.

* It fixes an issue where skip, sort, limit on collections that were on a different adapter weren't working correctly. (not a good idea because of memory issues but it works now)

I took @dmarcelino's awesome [PR for cross adapter joins](https://github.com/balderdashy/waterline-adapter-tests/pull/57) and expanded it a bit so I could run individual tests when debugging and they all passed for the adapters I ran it against. I'll submit that PR in a minute.

### DIFF
--- a/lib/waterline/query/finders/basic.js
+++ b/lib/waterline/query/finders/basic.js
@@ -308,10 +308,25 @@ module.exports = {
           criteria.joins.forEach(function(join) {
             if(!join.criteria) return;
             var c = normalize.criteria(join.criteria);
-            if(!c.sort) return;
-
             var alias = join.alias;
-            res[alias] = sorter(res[alias], c.sort);
+            if(c.sort) {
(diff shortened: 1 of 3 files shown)


dmarcelino:
@particlebanana, I'm astonished by the fact this didn't need too many lines of code. I'm no integrator expert but all your changes look good. I've ran tested these changes against my original cross-adapter tests and all passed. Great stuff!! :white_check_mark:

devinivy:
I dig it :+1:

dmarcelino:
Great to see this one merged! :smiley:

Issue #366, opened 2014-04-04
Association .populate() only populates defined attributes


danielbaak:
When using adapters that support schemaless mode (i.e. `schema: false`) populating a One-to-many association only populates the attributes defined in the associated model being populated. 

I'm pretty sure this is undesired behaviour?

particlebanana:
Submitted https://github.com/balderdashy/waterline/pull/985 which should be a fix for this but needs a bit of testing.

dmarcelino:
Hey @particlebanana, I think tests for this feature would be great candidates to be placed in a `features/schemaless` folder and enabled via adapter package.json like `"features": ["schemaless"]`. More about this idea on PR balderdashy/waterline-adapter-tests#64.

devinivy:
This helps me understand my questions in https://github.com/balderdashy/waterline/pull/985#issuecomment-102218703

Perhaps the `select` option should be removed altogether with #985.

dmarcelino:
If it's removed will `find({select[]})` still work in schemaless mode? Or the `select` you mentioned is not related to this one?

devinivy:
It appears to me that this `select` isn't directly related to the `find` option.  We could scour the code to confirm that, but the easiest thing to do would be to try and use sails-memory in schemaless mode :P

dmarcelino:
>  the easiest thing to do would be to try and use sails-memory in schemaless mode :P

True :)

dmarcelino:
`select` tests added in PR balderdashy/waterline-adapter-tests#71.

dmarcelino:
Fixed by PR #985, please re-open if you are still experiencing this issue.

PR #985, opened 2015-05-01
Allow the integrator to work on values not in the schema

particlebanana:
Allows you to populate on nosql databases and not lose values.

Possible fix for #366 but needs some more testing before merging in. Passes all current test but I didn't add any new tests for this case.

### DIFF
--- a/lib/waterline/model/lib/model.js
+++ b/lib/waterline/model/lib/model.js
@@ -59,16 +59,21 @@ var Model = module.exports = function(attrs, options) {
    * @return {String} output when this model is util.inspect()ed
    * (usually with console.log())
    */
-  this.inspect = function() {
-    var output;
-    try {
-      output = self.toObject();
+
+  Object.defineProperty(this, 'inspect', {
+    enumerable: false,
+    configurable: false,
+    writable: false,
+    value: function() {
+      var output;
+      try {
+        output = self.toObject();
+      }
+      catch (e) {}
+
+      return output ? util.inspect(output) : self;
     }
-    catch (e) {}
-
-    return output ? util.inspect(output) : self;
-  };
-
+  });
 
   return this;
 };
--- a/lib/waterline/query/integrator/index.js
+++ b/lib/waterline/query/integrator/index.js
@@ -66,8 +66,6 @@ module.exports = function integrate(cache, joinInstructions, primaryKey, cb) {
 
       var parentPK, fkToParent, fkToChild, childPK;
 
-      var childSelect;
-
       // N..N Association
       if ( instructions.length === 2 ) {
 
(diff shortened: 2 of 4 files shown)


dmarcelino:
@particlebanana, this seems to be breaking 2 unit tests [in travis](https://travis-ci.org/balderdashy/waterline/jobs/60908019#L1266):
* populate N..1 :: output should match the expected results
* populate N..M :: output should match the expected results

Is it the case where we need to update the tests?

particlebanana:
Yeah maybe, I'll have to look at those specific test and see what the deal is, they are unit tests. It passes all the adapter integration tests.

devinivy:
What do `.`-prefixed properties represent?

devinivy:
So, here's the deal.  I don't understand the integrator very well– I don't even understand how this PR fixes the schemaless issue.  But I do understand how this code change affected the unit tests.

The failing tests are at assertions [here](https://github.com/balderdashy/waterline/blob/master/test/unit/query/integrator.populate.js#L55) and [here](https://github.com/balderdashy/waterline/blob/master/test/unit/query/integrator.populate.js#L104).

It appears that the number of periods (`.`) prefixing an attribute signifies how deeply nested it is.  At this point in the integrator, nested records appear much like two joined sequel tables, with both parent and child columns appearing in the same result row.  Columns living on the child are prefixed with a `.`, and columns from the parent are not prefixed.

The integrator used to apply a `select` statement to select only some required fields off each child record (which appears, as stated before, with both parent and child columns in the same object).

With @particlebanana's change, rather than picking out `select`ed properties on a child record, the `select` option is ignored, and only non-parent columns are picked out to be used on the child records.  The non-parent columns are identified as not being prefixed with a `.`.

The tests expect the child records to contain both parent and child columns, as seen in the expected values here: https://github.com/balderdashy/waterline/blob/master/test/support/fixtures/integrator/populateResults.js#L14-L20.  Instead, only the period-prefixed properties appear in the results.

There don't appear to be tests of populate using the `select` option.  My questions are:

1. Is this behavior of omitting parent columns on child records okay at this point of the integrator?
2. Why is the `select` option now ignored?
3. How does this support schemaless models better than before?

particlebanana:
Made some more progress on this. However I found an issue with the operation building when m:m joins are used in a cross adapter query. Working that one out still. Looks like it's failing when the join table is on a different connection than the child.

@devinivy to answer your questions:

> Is this behavior of omitting parent columns on child records okay at this point of the integrator?

Yup we are cloning the `associatedChildRows` object. It's the same behavior as before just without the select filter.

> Why is the select option now ignored?

I removed all notion of `select` in the integrator. It was there from a previous time I guess. At this point in the query lifecycle all the queries have been completed so select would serve no purpose here anyway. In order for select to be useful it would need to be interpreted when the query was run. At this particular time the select was synonymous with the schema.

 > How does this support schemaless models better than before?

Before the select was pretty much always equal to the schema, so by only grabbing the `select` keys it wouldn't allow you to have values outside the schema. By returning everything, the schema/schemaless filter should kick in and filter out any results that don't match the schema if you have `schema: true` set.

particlebanana:
Will open a separate issue for the m:m join issue.

dmarcelino:
Good stuff @particlebanana, I'm working on enable features testing on all official adapter so we can wrap up from the testing side.

PR #986, opened 2015-05-03
createEach: use the output of async.map

dmarcelino:
instead of using helper array. to ensure the order is correct.

Supplement to #980, ensuring it passes the *.create() should return rows in the correct order when creating multiple rows* test.

cc: @devinivy, @kevinburkeshyp, @particlebanana, @brandonsimpson

### DIFF
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -45,30 +45,14 @@ module.exports = {
 
     var errStr = _validateValues(_.cloneDeep(valuesList));
     if(errStr) return usageError(errStr, usage, cb);
-
-    var records = [];
-
-    function create(value, next) {
-
-      // Handle undefined values
-      if(value === undefined) {
-        setImmediate(function() {
-          return next();
-        });
-      }
-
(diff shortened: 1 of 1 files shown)


devinivy:
Looks good to me.  One question: are undefined values filtered out in case the array is sparse or something?  I don't see why any entries should be undefined.  (I'm wondering if that should throw an error.)  I suppose this question might not be part of this PR– I see that the code is just meant to work the same as it did before.

dmarcelino:
> I see that the code is just meant to work the same as it did before.

Precisely @devinivy.

Regarding the origin of filtering out `undefined` values I traced it to https://github.com/balderdashy/waterline/commit/ebcf0662d8f01d2f70d4ad2c88494536468e9010 and https://github.com/balderdashy/waterline/commit/3e0c18016f43bdada790dce2ebeaf614bfb8c947, not a 100% sure why though.

dmarcelino:
Merging this in order to fix the integration tests in waterline-adapter-tests but we should continue the discussion about `undefined` and we can continue it here, for now.

PR #995, opened 2015-05-07
Fix special attribute names - aka "type" + failling test

atiertant:
check https://github.com/balderdashy/waterline-schema/issues/22 bug

### DIFF
--- a/lib/waterline/core/index.js
+++ b/lib/waterline/core/index.js
@@ -101,7 +101,7 @@ Core._initialize = function(options) {
   this._transformer = new Transformer(schemaAttributes, this.waterline.schema);
 
   // Transform Schema
-  this._schema.schema = this._transformer.serialize(this._schema.schema);
+  this._schema.schema = this._transformer.serialize(this._schema.schema,'schema');
 
   // Build up a dictionary of which methods run on which connection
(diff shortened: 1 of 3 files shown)


dmarcelino:
@atiertant, is this the failing test for balderdashy/waterline-schema#22 / #788?

atiertant:
@dmarcelino yes this test insert 'type' key in transformer

dmarcelino:
Hey @atiertant, I don't think passing a keyword in the function the most elegant solution yet I can't think of any better fix without a significant re-write. This is probably the most efficient fix, I think it's good! :white_check_mark: 

@devinivy, what do you think?

dmarcelino:
@tjwebb, can you please take a look?

dmarcelino:
@devinivy, have you had a chance to look into this?

devinivy:
I think it's okay too.

atiertant:
Hi ! 
Is there something else i could do to correct this bug? 
If code is ok for you, couldn't you merge it?
or should i modify something in my code?

dmarcelino:
@particlebanana, can you take a look please? Both me and @devinivy are OK with this.

particlebanana:
Yeah this looks great. Closing https://github.com/balderdashy/waterline-schema/issues/22 as well

PR #1002, opened 2015-05-11
Fix callback being called multiple times when updating model with associations

chotiwat:
In `createBelongsTo()` of `lib/waterline/query/dql/update.js`, `valuesObject.associations.models` could be modified inside the iterator of `async.each()` function. This could lead to the final callback, which should be called only once, being called multiple times.

Here's an example of the effect of modifying an array being iterated by `async.each()`:
```js
var async = require('async');
var array = [1, 2, 3, 4, 5];
async.each(array, function(item, next) {
  process.nextTick(function() {
    console.log(item);
    var index = array.indexOf(item);
    array.splice(index, 1);
    next();
  });
}, function() { console.log('done'); });
```
will output:
```
1
2
3
done
4
done
5
done
```

### DIFF
--- a/test/unit/query/query.update.nested.js
+++ b/test/unit/query/query.update.nested.js
@@ -90,6 +90,9 @@ describe('Collection Query', function() {
             },
             nestedModel: {
               model: 'nested'
+            },
+            nestedModel2: {
+              model: 'nested'
             }
           }
         });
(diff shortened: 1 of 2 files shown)


dmarcelino:
Thanks @chotiwat! I suspect this will fix #751 (beforeValidate runs 3x on nested update).

Can you also add an automated test that fails without your change and passes afterwards? Also, is there any risk that your change can impact updates or nested updates?

devinivy:
Seems kosher to me.  Really nice catch, @chotiwat .

chotiwat:
@dmarcelino I don't know if I put it at the right place, but I added a new test case that fails without this pull request. The test case requires asynchronous adapter though.
I also updated an existing test case for synchronous adapter that fails in another manner: the callback will be called only once but some nested attributes won't be processed.
This fix should not break anything, theoretically at least.

@devinivy Well thank you. :)

test/unit/query/query.update.nested.js
```diff
@@ -139,11 +142,100 @@ describe('Collection Query', function() {
...
+          if(err) return done(err);
+          query = colls.collections.user;
+          done();
+        });
+      });
+
+      it('should call back only once and reduce the newly created nested object down to two foreign keys', function(done) {
+        var count = 0;
+        query.update({}, { name: 'foo', nestedModel: { name: 'joe' }, nestedModel2: { name: 'jane' } }, function(err, status) {
+          assert(++count === 1);
```
devinivy: What does this test?  I don't see `count` used anywhere aside from being initially assigned `0`.

chotiwat: It is incremented and asserted in the callback (two lines below). If the callback is called more than once, the assertion will fail.

devinivy: Ah, I see!  All good.

dmarcelino:
The tests look good, thanks @chotiwat!

Issue #996, opened 2015-05-07
broken buffer for binary data in alter migration


atiertant:
adapter receive a broken buffer in createEach function for binary data while migrating in alter mode, in https://github.com/balderdashy/waterline/blob/master/lib/waterline/adapter/sync/strategies/alter.js 
line 99 : 

    backupData = _.cloneDeep(existingData);

the original data is clone with lodash' cloneDeep,this is working for many value but not for buffer : 
https://github.com/lodash/lodash/issues/311

this line should be change by : 

    backupData = _.cloneDeep(existingData, function dealWithBuffers(val) {
      if (val instanceof Buffer) {
        return val.slice();
      }
    });

devinivy:
Sounds like the right idea to me.  Feel free to submit a PR!

atiertant:
@devinivy in which file/folder of test do you think this kind of tests should be ?

devinivy:
Good question!  There are adapter integration tests for that here: https://github.com/balderdashy/waterline-adapter-tests/tree/1d190d5c664f803bbe02fc160955d518c019f21e/interfaces/migratable

dmarcelino:
Yeah, migratable sounds like a good place for it.

atiertant:
@devinivy, @dmarcelino  why adding this test in waterline-adapter-tests ? 
this is a waterline bug there is nothing that adapters developers can do... 
waterline test shouldn't unsure that passed value to adapters are correct ?

devinivy:
You can create a unit test for this too.  Here are the unit tests: https://github.com/balderdashy/waterline/tree/master/test/unit

They generally mirror the directory structure of the waterline project itself, so you'll have to create the `adapter` folder in there to place the unit test.

atiertant:
@devinivy what do you think of it ?

dmarcelino:
Fixed via #1005. Thanks @atiertant and @khalilTN!

PR #1005, opened 2015-05-13
Correction of alter strategy binary type bug + test

khalilTN:
When activating the "Alter" mode, waterline retrieve data, delete tables, recreate them and then reinsert data, if data contains buffers, executing cloneDeep() de lodash causes buffer to be converted to another type of object and then data loss. In this PR, correction of the bug and a test of the mode "alter".


atiertant:
fix https://github.com/balderdashy/waterline/issues/996

dmarcelino:
First impressions: the indentation is a bit off, most (all?) of waterline files use 2 spaces for tab size. Can you please fix it?

test/unit/adapter/strategy.alter.js
```diff
@@ -0,0 +1,139 @@
...
+            }
+        };
+        connections = {
+            'test_alter': {
+                adapter: 'fake'
+            }
+        };
+        adapters = {fake: adapter};
+    });
+    it('should recover data', function (done) {this.timeout(10000);
```
dmarcelino: Style alert, can `this.timeout(10000);` be moved to the line below?

khalilTN:
done

dmarcelino:
Thanks @khalilTN, I think this is good! :white_check_mark: 

@devinivy, what do you think?

dmarcelino:
@tjwebb, can you please take a look?

PR #1016, opened 2015-05-20
M:M cross adapter issue #1014

particlebanana:
Fix for https://github.com/balderdashy/waterline/issues/1014

Ensures that the join logic in operations gets set correctly on the connection registry.

I need to work on a major refactor of the operation builder so that it can be tested a lot better. Right now I'd be surprised if anyone besides me could figure out what the hell was going on! Now that I know how it works it can be refactored and tested. When I was building it there was a lot of ugghhhhh maybe this will work!


dmarcelino:
> Right now I'd be surprised if anyone besides me could figure out what the hell was going on!

You're probably right on that one! :blush:

PR #1018, opened 2015-05-21
Calling adapter's createEach method, fixes #1007

### LINKED ISSUE #1007: adapter.createEach not called
adapter.createEach is only called in alter mode to reinsert data.
shouldn't it optimize insertion of multiple data ?

masumsoft:
Sending createEach values to adapter's optimized .createEach() method
Will fallback to normal create if adapter doesn't implement .createEach()
This fixes #1007

### DIFF
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -51,8 +51,11 @@ module.exports = {
       return value !== undefined;
     });
     
+    //Sending to adapter's optimized createEach method
+    //Will fallback to normal create if adapter doesn't implement it.
+    self.adapter.createEach(filteredValues,cb);
     // Create will take care of cloning values so original isn't mutated
-    async.map(filteredValues, self.create.bind(self), cb);
+    //async.map(filteredValues, self.create.bind(self), cb);
   },
 
   /**
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -51,9 +51,65 @@ module.exports = {
       return value !== undefined;
     });
     
-    //Sending to adapter's optimized createEach method
-    //Will fallback to normal create if adapter doesn't implement it.
-    self.adapter.createEach(filteredValues,cb);
+    // Validate each record in the array and if all are valid
+    // pass the array to the adapter's createEach method
+    var validateItem = function(item, next) {
+      _validate.call(self, item, next);
+    }
+
+
+    async.each(filteredValues, validateItem, function(err) {
+      if(err) return cb(err);
+
+      // Transform Values
+      var transformedValues = [];
+
+      filteredValues.forEach(function(value) {
+
+        // Transform values
+        value = self._transformer.serialize(value);
+
+        // Clean attributes
+        value = self._schema.cleanValues(value);
+        transformedValues.push(value);
+      });
+
+      // Set values array to the transformed array
+      filteredValues = transformedValues;
+
+      // Pass attributes to adapter definition
+      self.adapter.createEach(filteredValues, function(err, values) {
+        if(err) return cb(err);
+
+        // Unserialize Values
+        var unserializedValues = [];
+
+        values.forEach(function(value) {
+          value = self._transformer.unserialize(value);
+          unserializedValues.push(value);
+        });
+
+        // Set values array to the transformed array
+        values = unserializedValues;
+
+        // Run AfterCreate Callbacks
+        async.each(values, function(item, next) {
+          callbacks.afterCreate(self, item, next);
+        }, function(err) {
+          if(err) return cb(err);
+
+          var models = [];
+
+          // Make each result an instance of model
+          values.forEach(function(value) {
+            models.push(new self._model(value));
+          });
+
+          cb(null, models);
+        });
+      });
+    });
+    
     // Create will take care of cloning values so original isn't mutated
     //async.map(filteredValues, self.create.bind(self), cb);
   },
--- a/lib/waterline/query/aggregate.js
+++ b/lib/waterline/query/aggregate.js
@@ -57,12 +57,14 @@ module.exports = {
     var processItem = function(item, next) {
       _process.call(self, item, next);
     }
+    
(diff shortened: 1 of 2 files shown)


lib/waterline/query/aggregate.js
```diff
@@ -51,8 +51,67 @@ module.exports = {
       return value !== undefined;
     });
     
+    // Validate each record in the array and if all are valid
+    // pass the array to the adapter's createEach method
+    var validateItem = function(item, next) {
+      _validate.call(self, item, next);
```
dmarcelino: Does this code do any extra validations in addition to `var errStr = _validateValues(_.cloneDeep(valuesList));` in [line 46](https://github.com/balderdashy/waterline/pull/1018/files#diff-762fb5e96b0c3cfe800c9025650adbb6R46)?

dmarcelino:
Hi @masumsoft, thanks for the PR! First impressions:
* In [create.js#L65-L76](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/create.js#L65-L76) each record goes through:
  ```javascript
    // Process Values
    var valuesObject = processValues.call(this, values);
  
    // Create any of the belongsTo associations and set the foreign key values
    createBelongsTo.call(this, valuesObject, function(err) {
      if(err) return cb(err);
  
      beforeCallbacks.call(self, valuesObject, function(err) {
        if(err) return cb(err);
        createValues.call(self, valuesObject, cb);
      });
    });
  ```
  From looking at your changes it's not clear to me how each record receives the same treatment before being handed to `adapter.createEach()`. Ideally we should use a DRY approach and call the same methods (`processValues()`, `createBelongsTo()`, `beforeCallbacks`).
* Related to the previous point, are `beforeUpdate` callbacks being called?
* Could you please add automated tests?

masumsoft:
@dmarcelino #1007 

Updated the PR with new commits. Now each record receives the same treatment like create.js before being handed to adapter.createEach(). But the methods were not directly called, rather duplicated into aggregate.js because create.js implementation ultimately calls the create method for each item, which is not desired in this case. 

I agree we should use a DRY approach and call the same methods (processValues(), createBelongsTo(), beforeCallbacks), but the methods are not exported to be used from other places. To export them, it would require to change the way how create.js works and I'm not confident enough to change the way it works as I'm new to waterline codebase. So by doing that, I fear breaking something else. 

I've also added a test case to validate if the adapter's createEach method was actually called. Hope this will be ok now.

lib/waterline/query/aggregate.js
```diff
@@ -51,8 +53,49 @@ module.exports = {
       return value !== undefined;
     });
     
-    // Create will take care of cloning values so original isn't mutated
-    async.map(filteredValues, self.create.bind(self), cb);
+    // process each item, their associations and beforeCreate callbacks
```
dmarcelino: Shouldn't we do this **only** if the adapter implements createEach? Because otherwise the `.create` will take care of it, right? I may be wrong but I get the impression that for adapters that don't implement `createEach` the values are being processed twice...

masumsoft: I think we don't need to check for it again here, as the check is already in place inside the adapter handler: https://github.com/masumsoft/waterline/blob/createeach-fix/lib/waterline/adapter/aggregateQueries.js#L25

The call to self.adapter.createEach() method in my implementation actually calls the underlined createEach() method inside adapter/aggregateQueries.js, which already handles if the adapter implements createEach or not.

dmarcelino: > The call to `self.adapter.createEach()` method in my implementation actually calls the underlined createEach() method inside adapter/aggregateQueries.js, which already handles if the adapter implements createEach or not.

This is why I believe the values are being processed twice when the adapter doesn't implement `.createEach()`. The values are being processed here once and then again inside [waterline's `.createEach()`](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/dql/create.js#L65-L76). This will make performance even worse for adapters not implementing `.createEach()`... I haven't yet tested the change but that's what it looks like....

test/unit/query/query.createEach.adapterMethod.js
```diff
@@ -0,0 +1,78 @@
...
+        waterline.initialize({ adapters: { foobar: adapterDef }, connections: connections }, function(err, colls) {
+          if(err) return done(err);
+          query = colls.collections.user;
+          done();
+        });
+      });
+
+
+      it('should call adapters createEach method', function(done) {
+        query.createEach([{ name: 'bob' }, { name: 'foo'}], function(err, values) {
```
dmarcelino: Calling `.createEach()` directly [is deprecated](https://github.com/balderdashy/waterline#query-methods), the test should call `.create([])`, which internally will call `.createEach()`.

masumsoft:
@dmarcelino I just updated the code with a check if adapter implements createEach, if not, create is being called. Also updated the test case as per your suggestion.

dmarcelino:
Hey @masumsoft, it's looking better! We still need to address the *DRYness* question. I'm not too familiar with this part of the code so I'll need some time to look into it. Meanwhile maybe @devinivy can lend a hand as he knows waterline internals better than me.

devinivy:
Yeah, the DRYness issue is really worth fixing now.  Maybe it's time for this lifecycle stuff to be broken out into its own part of the codebase.  Consider creating `lib/waterline/query/lifecycle/create.js` to place this creation lifecycle routine, then simplifying both `lib/waterline/query/dql/create.js` and `lib/waterline/query/aggregate.js`.  We can make issues to migrate other lifecycle happenings (e.g. update) into `lib/waterline/query/lifecycle/` as well.

particlebanana:
Hey @masumsoft I really appreciate the work done here. As discussed in a few places the hard part of this is getting the same logic run for each record. If we didn't have nested records it would be a simple top level array and all the writes could happens at once in a single update.

The main issue is the [nested create/updates](https://github.com/balderdashy/waterline/issues/1007#issuecomment-145177976). This is on the roadmap but will take some more work to get there.

PR #1021, opened 2015-05-21
Fix schema false migration

### LINKED ISSUE #966: When using Sails-disk and the default schema false, migrate alter will remove attributes not in the schema

particlebanana:
Fixes #966 and adds a few more simple alter migration tests


dmarcelino:
Looks good! :white_check_mark:

PR #1022, opened 2015-05-21
M m leapset fix

particlebanana:
If I understand #928 correctly then this should give the same result without the need for the transformation.

### DIFF
--- a/lib/waterline/query/dql/destroy.js
+++ b/lib/waterline/query/dql/destroy.js
@@ -74,6 +74,8 @@ module.exports = function(criteria, cb) {
         break;
       }
 
+      var transformedValues = _.map(result, self._transformer.unserialize.bind(self._transformer))
+
       function destroyJoinTableRecords(item, next) {
         var collection = self.waterline.collections[item];
         var refKey;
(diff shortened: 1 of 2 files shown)


dmarcelino:
I like this! :white_check_mark:

lib/waterline/query/dql/destroy.js
```diff
@@ -88,7 +93,14 @@ module.exports = function(criteria, cb) {
         // than crashing.
         if(!refKey) return next();
 
-        var mappedValues = result.map(function(vals) { return vals[pk]; });
+        // Make sure we don't return any undefined pks
+        var mappedValues = result.reduce(function(memo, vals) {
+          if (vals[pk] !== undefined) {
```
dmarcelino: Are undefined pks legitimate?

dmarcelino:
One question that @devinivy has asked in https://github.com/balderdashy/waterline/pull/928#issuecomment-93990300: is it legitimate to have undefined pas? Or should waterline throw an error there?

particlebanana:
I can't think of a situation where you would get a record back that didn't have a primary key. The only way this should happen would be if you were using `columnName` in the `map` which this PR should fix.

Anyone have any ideas how/why it would happen?

lib/waterline/query/dql/destroy.js
tjwebb: No

dmarcelino: So we probably don't need this code, I guess..

esopian:
Looks good to me.  :+1:

PR #1023, opened 2015-05-22
Nested create/updates with a custom pk fix

### LINKED ISSUE #519: Custom PK breaks association create
I'm using sails master and sails-mongo.

It seems like setting a custom primary key breaks creation of associations. This behaviour is a bit odd to me anyway.

### Case
I have 2 models. User and Visitor. User has a visitor attribute with model: 'visitor' and Visitor has a user attribute with mode: 'user'. Pretty basic oneToOne right there. For the rest, the Visitor model's username is the primaryKey.

Now, I try doing the following:

```js
var newUser = {
  email: 'some@email.com',
  password: 'super secure',
  visitor: {
    username: 'my username',
    credits: 1234
  }
};

User.create(newUser).exec();
```

#### Expected
A new User, and Visitor record associated to each other.

#### Actual
A new User, and no Visitor at all.

### A slight change
If I remove the primary key from username, it's still odd. It then creates the User record and the Visitor record but doesn't associate them to each other. I have to then manually update the new Visitor to associate it with its User record.

### Bonus question
How can I force blueprints to do a findOne if the primaryKey is something else?

I have the following as a route:

```js
module.exports.routes = {
  'get /visitor/:username': 'VisitorController.find'
}
``` 
but it always returns an array. I'd like it to do a findOne in stead. I tried changing find to findOne, but that didn't do much.

particlebanana:
This should fix #519 and #675. The issue was that when doing nested operations if the primaryKey was used then it assumed that the record already existed and would not try and create it. This was obviously a problem when you have non auto-incrementing primary keys where you need to specify the value.

Nested operations aren't my favorite thing but we should make what's there work. So what this does is check if the attribute is not auto-incrementing and that default autoPK flag is set to false and it's a create  or update method. If so it will let nested updates do it's thing. Otherwise it works as before and will just link the records together.

If someone else wouldn't mind giving this a quick test that would be great. @RWOverdijk would be great to test on mongo!

### DIFF
--- a/lib/waterline/model/lib/associationMethods/add.js
+++ b/lib/waterline/model/lib/associationMethods/add.js
@@ -127,7 +127,7 @@ Add.prototype.createAssociations = function(key, records, cb) {
       // If a custom PK was used on the associated collection and it's not
       // autoIncrementing, create the record. This allows nested
       // creates to work when custom PK's are used.
-      if(!relatedPK.autoIncrement && !associatedCollection.autoPK) {
+      if(!relatedPK || !relatedPK.autoIncrement && !associatedCollection.autoPK) {
         return self.createNewRecord(associatedCollection, schema, association, next);
       }
 
--- a/lib/waterline/model/lib/associationMethods/add.js
+++ b/lib/waterline/model/lib/associationMethods/add.js
@@ -114,6 +114,7 @@ Add.prototype.createAssociations = function(key, records, cb) {
   var attribute = this.collection._attributes[key];
   var collectionName = attribute.collection.toLowerCase();
   var associatedCollection = this.collection.waterline.collections[collectionName];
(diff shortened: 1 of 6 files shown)


dmarcelino:
@particlebanana, you're one fire :fire: !

It would be great to add a couple of tests for this in waterline-adapter-test so we could test against all official adapters.

particlebanana:
@dmarcelino I'll take a swing at that - we can hold off on merging this until it's tested.

dmarcelino:
BTW, did these changes pass all new tests added by balderdashy/waterline-adapter-tests#74?

particlebanana:
@dmarcelino yup, re-running them now and with this merged they should all pass.

dmarcelino:
Awesome work! :smiley:

Issue #1035, opened 2015-05-29
afterUpdate on create


RWOverdijk:
https://github.com/RWOverdijk/unexpectedUpdate

1. mysql > `create database testcase`
2. npm i
3. sails console
4. The output below (coming from [here](https://github.com/RWOverdijk/unexpectedUpdate/blob/master/api/models/Foo.js#L20)

```bash
sails> Foo.create({bar: {a:'b'}, bat: [{a:'b'}]}).then(console.log).catch(console.error);
{ _bitField: 0,
  _fulfillmentHandler0: undefined,
  _rejectionHandler0: undefined,
  _progressHandler0: undefined,
  _promise0: undefined,
  _receiver0: undefined,
  _settledValue: undefined }
sails> This makes me sad { id: 1,
  createdAt: Fri May 29 2015 16:45:10 GMT+0200 (CEST),
  updatedAt: Fri May 29 2015 16:45:10 GMT+0200 (CEST),
  bar: 1 }
{ bar: 1,
  createdAt: Fri May 29 2015 16:45:10 GMT+0200 (CEST),
  updatedAt: Fri May 29 2015 16:45:10 GMT+0200 (CEST),
  id: 1 }
```

It updates Foo. I'm doing a create. I've tried adding `via`, that's not the issue.

I've traced the update back to [here](https://github.com/balderdashy/waterline/blob/master/lib/waterline/model/lib/defaultMethods/save.js#L73) (which was a nightmare to find).

I don't expect an update at all. Especially not with all values.

On top of that, there's this beauty of a query log:

```sql
INSERT INTO `bar` (`a`, `createdAt`, `updatedAt`) values ('b', '2015-05-29 16:48:57', '2015-05-29 16:48:57')
INSERT INTO `foo` (`bar`, `createdAt`, `updatedAt`) values (1, '2015-05-29 16:48:57', '2015-05-29 16:48:57')
SELECT `foo`.`id`, `foo`.`createdAt`, `foo`.`updatedAt`, `foo`.`bar` FROM `foo` AS `foo`  WHERE `foo`.`id` = 1
UPDATE `foo` AS `foo` SET `bar` = 1, `createdAt` = '2015-05-29 16:48:57', `updatedAt` = '2015-05-29 16:48:57', `id` = 1  WHERE `foo`.`id` = 1
[...]

PR #1044, opened 2015-06-05
[wip] Accept empty save() calls

RWOverdijk:
The tests fail. However, this does work for me. I know this solution is wrong, I just really need some feedback on this one.

### DIFF
--- a/lib/waterline/model/lib/associationMethods/update.js
+++ b/lib/waterline/model/lib/associationMethods/update.js
@@ -37,7 +37,7 @@ var Update = module.exports = function(collection, proto, mutatedModels, cb) {
 
   // Clone values so they can be mutated
   var _values = _.cloneDeep(values);
-  
+
   // For any nested model associations (objects not collection arrays) that were not changed,
   // lets set the value to just the foreign key so that an update query is not performed on the
   // associatied model.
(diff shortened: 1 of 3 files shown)


lib/waterline/model/lib/associationMethods/update.js
```diff
@@ -65,7 +65,11 @@ var Update = module.exports = function(collection, proto, mutatedModels, cb) {
     var reduced = nestedOperations.reduceAssociations(collection.identity, collection.waterline.schema, vals);
     _values = _.merge(_values, reduced);
   });
-  
+
+  if (!mutatedModels.length) {
+    return cb(null, proto);
```
dmarcelino: Why `proto`? Usually update returns an array of values and if the criteria did not match any record it returns an empty array.

RWOverdijk: This is just to ignite a discussion. This is wrong to begin with, because simple updates don't have mutatedModels.

devinivy:
What specifically are you trying to achieve?  I assume the end goal is to avoid an update during the nested create routine if it's not necessary, based upon #1035 .

RWOverdijk:
@devinivy Correct. Save should be called, but not trigger update if there are no changes. So basically, only process what's been altered (in this case, the associations).

devinivy:
What is your specific plan to fix that?  Currently it looks like you're trying to tackle it from within `save` rather than within the nested create routine.

If that's true, I think the main way to do this is to detect if all non-collection attributes do not need updating.  If they don't need updating, you can return `[proto]` and skip the `collection.update`.  The only problem with what you have is that it's skipping `collection.update` if _some_ non-collection attributes (the attributes that represent foreign keys) do not need updating.

Issue #1001, opened 2015-05-11
Setting stack on WLError causes infinite loop


t3rminus:
Seems to have been introduced with the recent fix for this issue: https://github.com/balderdashy/waterline/issues/913

WLError.stack is set to itself in the setter, causing an infinite loop
```javascript
Object.defineProperties(WLError.prototype, {
  stack: {
    enumerable: true,
    get: function () {
      return util.format('Error (%s) :: %s\n%s', this.code, this.reason, this.rawStack);
    },
    set: function (value) {
      this.stack = value; // Infinite loop?
    }
  }
/// Snip
});
```

When I cause a WLValidationError, the server crashes with this error:
```
sails/node_modules/waterline/lib/waterline/error/WLError.js:129
verbose:     set: function (value) {
                  ^
verbose: RangeError: Maximum call stack size exceeded
    at WLValidationError.Object.defineProperties.stack.set [as stack] (sails/node_modules/waterline/lib/waterline/error/WLError.js:129:19)
    at WLValidationError.Object.defineProperties.stack.set [as stack] (sails/node_modules/waterline/lib/waterline/error/WLError.js:130:18)
    at WLValidationError.Object.defineProperties.stack.set [as stack] (sails/node_modules/waterline/lib/waterline/error/WLError.js:130:18)
    at WLValidationError.Object.defineProperties.stack.set [as stack] (sails/node_modules/waterline/lib/waterline/error/WLError.js:130:18)
(... snip about 50x more ...)
```

dmarcelino:
@t3rminus, |I take it you're referring to the changes in [PR #920](https://github.com/balderdashy/waterline/pull/920/files#diff-8529b4b9f350440863f04bae69842609R123)?

@tjwebb, can you comment?

eddieajau:
I have run into this bug too. I accidentally passed an undefined variable to a `findOne` call on a collection (albeit in a Sails test harness).

dmarcelino:
PR #1045 should fix this. @t3rminus, @eddieajau can you please try it?

PR #1045, opened 2015-06-06
prevents infinite recursion in "WLError.stack" setter

shaunc:


dmarcelino:
Hi @shaunc, thanks for this. I'm assuming this is the fix for #1001?

shaunc:
yes ... I am only assuming that "rawStack" is what should be set -- this "works for me" but don't know what users of setters really expect. (OTOH not documented so why not this? :))

dmarcelino:
@tjwebb, can you give a glance at this? This fixes #1001 which was caused by PR #920.

>  (OTOH not documented so why not this? :))

Very true :)

dmarcelino:
@devinivy, what do you think?

devinivy:
I think it's fine!

dmarcelino:
Closing this over PR #1061 which includes @shaunc's fix plus a failing test.

Issue #1015, opened 2015-05-20
Unhandled rejection Error


anasyb:
Have been getting this error a bunch of times a day since a week or so. Have no idea why or from where is it originating in the async spaghetti. Spent couple of days trying to pinpoint it with no much luck.

The only difference I get is that the number of `[object Object]` listed varies, sometimes its one, sometimes its a bunch of them... or in this case its just 2.

Any ideas/hints guys on what would be the cuz of this and why the number of `[object Object]` is different each time?

```
Unhandled rejection Error: [object Object],[object Object]
    at Object.ensureErrorObject (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/bluebird/js/main/util.js:226:20)
    at Promise._rejectCallback (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/bluebird/js/main/promise.js:416:22)
    at Object.reject (/usr/local/lib/node_modules/sails/node_modules/waterline/node_modules/bluebird/js/main/promise.js:433:17)
    at /usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/model/lib/defaultMethods/save.js:165:18
    at /usr/local/lib/node_modules/sails/node_modules/async/lib/async.js:454:17
    at /usr/local/lib/node_modules/sails/node_modules/async/lib/async.js:444:17
    at Array.forEach (native)
    at _each (/usr/local/lib/node_modules/sails/node_modules/async/lib/async.js:46:24)
    at Immediate.taskComplete (/usr/local/lib/node_modules/sails/node_modules/async/lib/async.js:443:13)
    at processImmediate [as _immediateCallback] (timers.js:369:17)
```

dmarcelino:
Hi @anasyb, the only waterline related line I see in the stack trace is:
``` shell
    at /usr/local/lib/node_modules/sails/node_modules/waterline/lib/waterline/model/lib/defaultMethods/save.js:165:18
```

Which refers to [defaultMethods/save.js#L165](https://github.com/balderdashy/waterline/blob/master/lib/waterline/model/lib/defaultMethods/save.js#L165):
```javascript
        deferred.reject(failedTransactions);
```

I believe this code runs when you do `someModel.save(cb)`. Is it possible you're not processing/catching the result of one of those `.save()` operations?

anasyb:
Thanks @dmarcelino, a quick find for all `.save(` revealed that this is the case... the error there is simple:
[...]

PR #1048, opened 2015-06-09
Hotfix/update on create

RWOverdijk:
This PR makes it so that:

* You can view the diff on a model
* You don't get afterUpdate to trigger on create
* You don't get an update at all, on create (because there's no diff)
* .save() only updates changed fields, in stead of everything
* There's no association cache anymore
* The original values are retrievable on the model.

### Notes to self
* ~~`.compareModelValues()` in `save.js` can probably go.~~
* ~~`.findPrimaryKey()` can probably go. Context has a primaryKey property.~~
* ~~Don't forget to set status back to clean after various save operations.~~

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -170,8 +170,8 @@ module.exports = function(context, proto, cb) {
     // Rebuild proper criteria object from the original query
     var PK = context.primaryKey;
 
-    if(!Array.isArray(results.updateRecord)) {
-      if (results.updateRecord) {
+    if(!results.updateRecord.length) {
+      if (!Array.isArray(results.updateRecord)) {
         return cb (null, results.updateRecord);
       }
 
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -21,7 +21,7 @@ var noop = function() {};
  * @api public
  */
 
-module.exports = function(context, proto, cb, associationsOnly) {
+module.exports = function(context, proto, cb) {
 
   var deferred;
 
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -31,55 +31,32 @@ module.exports = function(context, proto, cb) {
 
   cb = cb || noop;
 
+  var dirtyProperties = proto.dirty(),
+      mutatedModels   = [];
+
+  _.forIn(proto.associations, function (value, key) {
+    if (dirtyProperties[key]) {
+      mutatedModels.push(key);
+    }
+  });
+
   /**
-   * TO-DO:
-   * This should all be wrapped in a transaction. That's coming next but for the meantime
-   * just hope we don't get in a nasty state where the operation fails!
+   * @todo This should all be wrapped in a transaction. That's coming next but for the meantime
+   * @todo just hope we don't get in a nasty state where the operation fails!
    */
-
-  var mutatedModels = [];
-
   async.auto({
+    updateRecord: function(next) {
+      if (false === dirtyProperties) {
+        return next();
+      }
 
-    // Compare any populated model values to their current state.
-    // If they have been mutated then the values will need to be synced.
-    compareModelValues: function(next) {
-      var modelKeys = Object.keys(proto.associationsCache);
-
-      async.each(modelKeys, function(key, nextKey) {
-
-        if(!hop(proto, key) || proto[key] === undefined) return nextKey();
-
-        var currentVal = proto[key];
-        var previousVal = proto.associationsCache[key];
-
-        // Normalize previousVal to an object
-        if(Array.isArray(previousVal)) previousVal = previousVal[0];
-
-        if(deep(currentVal, previousVal)) {
-          mutatedModels.push(key);
-        }
-
-        nextKey();
-      }, next);
+      new updateInstance(context, dirtyProperties, mutatedModels, next);
     },
-
-    // Update The Current Record
-    updateRecord: ['compareModelValues', function(next) {
-
-      // Shallow clone proto.toObject() to remove all the functions
-      var data = _.clone(proto.toObject());
-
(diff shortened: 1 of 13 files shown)


lib/waterline/model/lib/defaultMethods/save.js
```diff
@@ -170,15 +146,26 @@ module.exports = function(context, proto, cb) {
...
       return cb(error);
     }
 
-    var obj = results.updateRecord[0].toObject();
+    if (dirtyProperties) {
+      proto.clean();
+    }
+
+    if (typeof results.updateRecord === 'undefined') {
+      results.updateRecord = proto;
```
RWOverdijk: Not sure about this. What should be done here?

RWOverdijk:
Tests currently fail. The first two errors are because of [this piece of code](https://github.com/balderdashy/waterline/blob/master/test/unit/model/association.add.hasMany.id.js#L31). The model didn't change. This PR is now responsible for stopping .save() from updating the model if nothing changed. So, .updated is empty... So you get this error. This means that the reason these tests fail, is because they relied on .update() being called even if there weren't any changes.

TL;DR; I will change the tests to make sure they reflect the truth.

As to the rest of the tests, I'll fix those after.

RWOverdijk:
Now that the build is succeeding and all tests for the changes have been written... Let's talk. It's a pretty massive change.

RWOverdijk:
Now that the build is succeeding and all tests for the changes have been written... Let's talk. It's a pretty massive change.

dmarcelino:
Nice... I'll need some time to digest this, meanwhile lets cc @devinivy, @particlebanana and @tjwebb.

RWOverdijk:
Alrighty. I can confirm this solves my initial issue by the way. afterUpdate isn't being called on create anymore.

dmarcelino:
So, for reference, this replaces PR #710, and fixes #1035 and #683, right?

RWOverdijk:
@dmarcelino Correct. As well as reduce the number of queries made most of the time and removing unneeded code (mutatedModels, associationCache). But it also no longer calls afterUpdate on create so, technically it's a BC break.

dmarcelino:
So many good things, it must be an early xmas! :christmas_tree:

> But it also no longer calls afterUpdate on create so, technically it's a BC break.

I'm not sure it qualifies as breaking change since afterUpdate on create is a bug and not an intended (or even documented) feature. In other words, I doubt people would code expecting that behaviour, but we should discuss it further.

RWOverdijk:
I've also added .getPrimaryKey, but that can be taken out.

Pretty sure that reserving .clean and .dirty on models should be considered a BC break though haha. People might be using those.

dmarcelino:
You're right, I'm pretty sure they are. We should should think of alternatives that will not colide with user's models. Something like `isCleanRecord()` and `isDirtyRecord()` should minimize collisions...

RWOverdijk:
@dmarcelino Probably .getDirtyProperties() and .setClean() then. isCleanRecord() currently equals to `.dirty() === false`. 

I personally like . unsynced() or .getUnsynced() in stead of .dirty(), and .synced() or .setSynced() for .clean(). But I took the naming from the other PR.

particlebanana:
@RWOverdijk awesome work! This is great. We definitely need to change the `.dirty` and `.clean` method names though. 

Initially I was thinking we could add underscores to represent private methods but having access to the dirty properties could be pretty useful. I'm not sure what the use case for having `.clean()` be public is though. 

Rails uses the `changed` wording so we could do something along those lines - `.getChanged()` would return the attributes that have changed then make `._resetSynced()` a private method. Not as friendly as `.dirty()` and `.clean()` but maybe less confusing? Just a thought.

RWOverdijk:
@particlebanana .getChanged() makes it implicit and simple, with fewer chance of naming collision. I like it. Perhaps use the same wording for .clean()? .applyChanges()? Or .submitState().. Something. It's not really a reset, because you're saying that the current state has been synced as is now the currentState. Something like .acceptCurrentState(properties)?

RWOverdijk:
cool cool. Needs rebasing now

dmarcelino:
> cool cool. Needs rebasing now

:disappointed:  We are closing `0.10` and preparing one last hotfix. Afterwards our focus will be in `0.11`

For reference, this should also fix #1015.

RWOverdijk:
@dmarcelino Just let me know when to rebase this stuff.

dmarcelino:
Hey @RWOverdijk, I believe your PR was broken by all the linting that took place, there are still a few linting rules breaking, https://github.com/balderdashy/waterline/pull/1084#issuecomment-121133140:

> screenshot from 2015-07-14 00 56 32

So there is some risk that further linting will again break your PR :confused: 

BTW, we forked `0.10` from master for hotfixes and master is now dedicated to `0.11`.

RWOverdijk:
@dmarcelino I'll sit patiently.

RWOverdijk:
@dmarcelino Come on man.

weyert:
Any news whether this PR gets merged into master any time soon?

devinivy:
I think it's safe to rebase this before linting continues.

RWOverdijk:
Does it even make sense at this point? Is it safe?

RWOverdijk:
Fuck it.

devinivy:
@RWOverdijk :fireworks:

saschagehlich:
@RWOverdijk Why did this never get merged?

RWOverdijk:
@saschagehlich Because I got tired of waiting and having to keep rebasing. I went full bender and built my [own ORM](https://wetland.spoonx.org). With blackjack, and hoo... Ladies.

RWOverdijk:
@saschagehlich There's a [sails hook](https://github.com/SpoonX/sails-hook-wetland) but I haven't tested it with 1.0 yet. I'm using it in production. :)

It doesn't work with document based storage. It's an ORM, not ODM. :)

saschagehlich:
NP, been having problems with MongoDB and CouchDB in larger setups anyways. Postgres ftw!

RWOverdijk:
Haha I agree. 👍  Let's not go too offtopic here. If you have questions you can hit me up on gitter. https://gitter.im/spoonx/dev

PR #1054, opened 2015-06-10
Added ability to transform columnName in populate() criteria

KeitaMoromizato:
Added ability to transform columnName in populate() criteria.

Before this fix, `where` statement in populate() cannot use model attribute name.

Like this case.
```
// Article.js
module.exports = {
  tableName: 'articles',

  attributes: {
    title: 'string',
    body: 'string',
    categories: {
      collection: 'category',
      through: 'articlecategory'
    },
    publicationDate : {
      columnName: 'publication_date',
      type: 'Date'
    }
  }
};

// Category.js
module.exports = {
  tableName: 'categories',

  attributes: {
    name: 'string'
    articles: {
      collection: 'article',
      through: 'articlecategory'
    }
  }
}

// ArticleCategory.js
module.exports = {
  tableName: 'articles_categories',
  tables: ['articles', 'categories'],
  junctionTable: true,

  attributes: {
    article: {
      columnName: 'article',
      type: 'integer',
      foreignKey: true,
      references: 'article',
      on: 'id',
      onKey: 'id',
      via: 'category'
    },
    category: {
      columnName: 'category',
      type: 'integer',
      foreignKey: true,
      references: 'category',
      on: 'id',
      via: 'article'
    }
  }
};
```

`where` statement in `populate()` cannot use attributeName defined at model.

```
// find category with articles
Category.find().populate('articles', where: {publication_date: {'<=': new Date()}}).then(function(categories) {
  // ...
}).catch(function(error) {});

// ERROR on `publicationDate`
Category.find().populate('articles', where: {publicationDate: {'<=': new Date()}}).then(function(categories) {
  // ...
}).catch(function(error) {});

```

After this fix, `where` statement in populate() works both `columnName` and `attributeName`.

And, I fixed test case `transformations.initialize.js`. The 2nd argument of `Transformer` expects simple object (like `{article: ..., category: ...}`).

### DIFF
--- a/test/unit/core/core.transformations/transformations.initialize.js
+++ b/test/unit/core/core.transformations/transformations.initialize.js
@@ -16,7 +16,7 @@ describe('Core Transformations', function() {
           }
         };
 
-        transformer = new Transformer(attributes, { models: {}, collections: {} });
+        transformer = new Transformer(attributes, {});
       });
 
       it('should set a username transformation', function() {
@@ -39,7 +39,7 @@ describe('Core Transformations', function() {
       it('should NOT set a username transformation', function() {
         var msg = (function() {
           try {
-            new Transformer(attributes, { models: {}, collections: {} });
+            new Transformer(attributes, {});
           } catch(e) {
             return e.message;
           }
--- a/lib/waterline/core/transformations.js
+++ b/lib/waterline/core/transformations.js
@@ -69,6 +69,8 @@ Transformation.prototype.initialize = function(_attributes, tables) {
   _mapping(_attributes, this._transformations);
 
   for (var tableName in tables) {
+    if (!tables[tableName].attributes) return;
+
     this._allTransformations[tableName] = {};
     _mapping(tables[tableName].attributes, this._allTransformations[tableName]);
   }
--- a/lib/waterline/core/transformations.js
+++ b/lib/waterline/core/transformations.js
@@ -20,7 +20,6 @@ var Transformation = module.exports = function(attributes, tables) {
 
   // Hold an internal mapping of keys to transform
   this._transformations = {};
-  this._allTransformations = {};
 
   // Initialize
   this.initialize(attributes, tables);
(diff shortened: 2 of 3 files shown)


devinivy:
Maybe I'm a little bit paranoid, but now doesn't each collection's transformer hold the transformations for every collection?  Why not use the child collection's transformer to make the transformation?

KeitaMoromizato:
Maybe collection doesn't have child collection's transformer. I could not find applicable code...

devinivy:
You would have to get the child collection name by diving into the schema, I will look for a code example.

devinivy:
It will look something like this: https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/joins.js#L185-L188

My only other concern as far as testing goes, is that this works for many-to-many and one-to-many scenarios.  Could that `join.child` ever reference a junction table?

KeitaMoromizato:
Thanks!
I will rewrite my code by reference to that.

I checked both many-to-many and one-to-many. many-to-many sample is above(`Article`,`Category` and `ArticleCategory`). I tried one-to-many senario like following.

```
// Article.js
module.exports = {
  tableName: 'articles',

  attributes: {
    title: 'string',
    body: 'string',
    images: {
      collection: 'ArticleImage',
      via: 'article'
    },
  }
}

// ArticleImage.js
module.exports = {
  tableName: 'article_images',

  attributes: {
    size: 'int',
    article: {
      columnName: 'article_id',
      model: 'Article',
      type: 'integer',
      required: true
    },
    fileName: {
      type: 'string',
      columnName: 'file_name',
    },
  }
}

// Query
Article.find({id: 1}).populate('images', {where: {fileName: "test.png"}}).then( ...

```

test/unit/core/core.transformations/transformations.initialize.js
```diff
@@ -16,7 +16,7 @@ describe('Core Transformations', function() {
           }
         };
 
-        transformer = new Transformer(attributes, { models: {}, collections: {} });
```
devinivy: Were these model and collection properties just found to be unused?

devinivy:
Awesome, thanks!  I just have one more question which I placed in code comments.  This will be great :)

test/unit/core/core.transformations/transformations.initialize.js
KeitaMoromizato: What does mean "model and collection properties"?
"model" and "collection" is just a object name? (same `{article: {...}, category: {...}}`)
I thought this test code means nested object like `{collections: {article: {...}, category: {...}}, models: {...}}`. So I fixed it.

Transformer's 2nd argument(table) is not used. So I cannot understand this argument.
But transformer initialized at `core/index.js` with `this.waterline.schema`. `this.waterline.schema` is simple object like `{article: {...}, category: {...}`. Maybe Transformer's 2nd argument doesn't expect nested object.

dmarcelino:
@KeitaMoromizato, would it be possible to add an automated test that fails before your fix is applied and works afterwards? The test can be based on the example you wrote on the first post. Thanks!

particlebanana:
This looks good, I'll whip up a quick test in (adapter-tests I guess?) so we have a before/after test to show it works.

particlebanana:
Added a failing unit test

dmarcelino:
@particlebanana, I assume the #1078 is passing after applying this change to waterline right? If so, I'm good with merging this. :white_check_mark:

particlebanana:
Awesome, test in https://github.com/balderdashy/waterline/pull/1078 failed and after merged this passes.

particlebanana:
Thanks @KeitaMoromizato!

dmarcelino:
Thank you @KeitaMoromizato!

dmarcelino:
> Awesome, test in #1078 failed and after merged this passes.

Cool!

Issue #1067, opened 2015-06-24
Implement type DOUBLE


SamProtas:
type: double is currently cast as 'string' as it's not officially supported.

https://github.com/balderdashy/waterline/blob/317cdf0173990face8afe67b336bc6e6d0fe5630/lib/waterline/core/typecast.js#L40

Many adapters implement double (sails-postgresql) or at least degrade gracefully (sails-mysql coerces to FLOAT).

Please add DOUBLE to types.js in order to allow adapters to deal with type DOUBLE in the best way for that database.

SamProtas:
PR submitted:

https://github.com/balderdashy/waterline/pull/1068

devinivy:
Hey @SamProtas, does this feature in the works look like it might do the trick? https://github.com/balderdashy/waterline/pull/981

SamProtas:
@devinivy I haven't delved to deep but that looks like a broader (inclusive) feature/fix. Is there any timeline for the implementation of #981?

My only plug for #1088 as a bandaid is that *double* is an extremely common field type that's already implemented in many (possibly all) of the most common adapters. It's a non-breaking change that we could merge today :+1: 

These 3 lines of code will given everyone access to type=double today!

I get if bandaids are frowned upon though....

devinivy:
@dmarcelino what do we know about the timeline on #981?  I do understand the desire have more granular typing.  To get your PR through we will have to make sure all the supported adapters do actually understand how to deal with that type.  Does javascript have enough bits per float to support what most DBs would consider a double?

dmarcelino:
> what do we know about the timeline on #981?

Still waiting on @particlebanana's input. Not sure if he had a chance to look at it. 

>  Does javascript have enough bits per float to support what most DBs would consider a double?

[javascript numbers](http://www.w3schools.com/js/js_numbers.asp) are 64-bit Floating Point so I think it would align well with DB `double`.

> I get if bandaids are frowned upon though....

I think this would open a worse precedent. Supporting `double` in waterline core would suggest all adapters can deal with it which I doubt is the case. I would prefer pushing  #981 as it's a definitive solution for the different types issues.

SamProtas:
Full disclosure, I haven't tested this yet. 

Two wrongs don't necessarily make a right, but to play devil's advocate, anchor doesn't appear to be able to handle `type: 'objectid'` and I don't see anything in waterline to deal with this either....

That said, +1 for #981

SamProtas:
[...]

PR #1068, opened 2015-06-25
Implements attribute Type=Double so adapters implementations work.

SamProtas:


dmarcelino:
For reference, this has been discussed in #1067 and the idea is that this should be supported by the adapters after PR #981 goes through in `0.11`.

particlebanana:
A few different ways of doing this. If anything the `dbType` seems like the solution when using auto migrations. I still really think all the db migration stuff lives outside of the core Waterline codebase.

Thanks for the input @SamProtas and I'll get you an officially supported method of doing this in a future version. For now checkout [Sails-Migrations](https://github.com/BlueHotDog/sails-migrations) for handling db changes in a more granular way.

Issue #1073, opened 2015-07-01
registerConnection's use of _.cloneDeep is causing ssl to fail


cludden:
When attempting to connect to a Google CloudSQL database using the sails-mysql adapter and a mysql with ssl connection definition:

```
module.exports.connections = {
    mysql: {
        adapter: 'sails-mysql',
        host: process.env.CLOUD_SQL_HOST,
        user: process.env.CLOUD_SQL_USER,
        database: process.env.CLOUD_SQL_DB,
        ssl: {
            key: fs.readFileSync(path.join(cloudsql, 'client-key.pem')),
            cert: fs.readFileSync(path.join(cloudsql, 'client-cert.pem')),
            ca: fs.readFileSync(path.join(cloudsql, 'server-ca.pem'))
        }
    },
```

the connection is failing with the following error:
```
error: Error (E_UNKNOWN) :: Encountered an unexpected error
: Could not connect to MySQL:
Error: unable to verify the first certificate
    at afterwards (/Users/cludden/Documents/Projects/wccsails2/node_modules/sails-mysql/lib/connections/spawn.js:72:13)
```

after much sifting, this seems to be due to the fact that the ssl file buffers returned from fs.readFileSync are being converted into js objects when cloned via the _.cloneDeep method:

adding a couple of log statements to waterline.js:
```
// lib/waterline.js

        // Call the registerConnection method
        console.log(config);
        console.log(_.cloneDeep(config));
        connection._adapter.registerConnection(_.cloneDeep(config), usedSchemas, function(err) {
          if(err) return nextItem(err);
          nextItem();
        });
```

yields the following:
```
{ pool: true,
  connectionLimit: 5,
  waitForConnections: true,
  adapter: 'sails-mysql',
  host: ********,
  user: ********,
  database: ********,
  ssl: 
   { key: <Buffer 2d 2d 2d //etc ... >,
     cert: <Buffer 2d 2d 2d //etc... >,
     ca: <Buffer 2d 2d 2d 2//etc ... > },
  identity: 'mysql' }
```

followed by:
```
{ pool: true,
  connectionLimit: 5,
  waitForConnections: true,
  adapter: 'sails-mysql',
  host: *******,
  user: *******,
  database: *******,
  ssl: 
   { key: 
      { '0': 45,
        '1': 45,
        '2': 45,
        '3': 45,
        '4': 45,
        '5': 66,
        '6': 69,
        '7': 71,
        '8': 73,
        '9': 78,
[...]

PR #1074, opened 2015-07-01
removed deepClone of config settings

cludden:
related to:
https://github.com/balderdashy/waterline/issues/1073

### DIFF
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -200,7 +200,7 @@ Waterline.prototype.initialize = function(options, cb) {
         });
 
         // Call the registerConnection method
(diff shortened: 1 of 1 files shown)


devinivy:
@tjwebb @particlebanana what're your thought on this?  The issue we're trying to fix is allowing buffers in config (for, say, SSL keys, etc).  But perhaps we don't need to clone config at all.

tjwebb:
Yea I think here `_.cloneDeep` might be causing more issues than it might potentially solve.

particlebanana:
passes all the tests and sounds like a reasonable assumption to me.

PR #1076, opened 2015-07-02
Return empty/zero from IN w/empty array

slester:
Calling Collection.destroy({id: []}) destroys *all* documents. [Functionality exists in normalize that returns false if there's an empty array](https://github.com/balderdashy/waterline/blob/317cdf0173990face8afe67b336bc6e6d0fe5630/lib/waterline/utils/normalize.js#L191), but false is OR'd with {}, [resulting in a {} being passed to the sails-mongo adapter](https://github.com/balderdashy/sails-mongo/blob/master/lib/adapter.js#L367), which deletes everything.

This PR just applies [what is already implemented for find()](https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/basic.js#L222).

We found this out the hard way. :(

### DIFF
--- a/lib/waterline/query/dql/count.js
+++ b/lib/waterline/query/dql/count.js
@@ -39,6 +39,12 @@ module.exports = function(criteria, options, cb) {
   // Normalize criteria and fold in options
   criteria = normalize.criteria(criteria);
 
+  // If there was something defined in the criteria that would return no results, don't even
+  // run the query and just return 0
+  if(criteria === false) {
+      return cb(null, 0);
+  }
+
   if(_.isObject(options) && _.isObject(criteria)) {
     criteria = _.extend({}, criteria, options);
   }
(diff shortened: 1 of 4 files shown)


dmarcelino:
This seems reasonable to me, it makes `destroy` and `find` consistent, and protects users from deleting all their records accidentally :white_check_mark: 

@slester, I would love if this PR included an automated test... :)

@devinivy, what are your thoughts?

devinivy:
Just to make sure I understand, how would `Collection.destroy()` behave?

slester:
Collection.destroy({id: []}) should return a callback with no error but also an empty "deleted" array. Right now, it behaves exactly like Collection.destroy({}).

slester:
@dmarcelino Added a unit test!

dmarcelino:
Cool! :grinning:

devinivy:
@slester what would `Collection.destroy()` do, exactly as typed?

slester:
`Collection.destroy()` will do the same thing as `Collection.destroy({})`, namely delete everything in the collection. That behavior won't be changed in our code.

devinivy:
Cool– sounds good :)

slester:
Let me know if anything is lacking for this PR!

tjwebb:
+1

dmarcelino:
Thanks @slester!

Issue #824, opened 2015-02-04
Use Joi for validation


mattapperson:
Waterline is a great project, but the validation in models is a little weak. Had or would consideration be given to moving to https://github.com/hapijs/joi for this aspect of the project?

particlebanana:
I've briefly looked into Joi as a replacement for Anchor but haven't evaluated how much effort it would involve or if the api would still work as expected. Would love to get it moved over though!

mattapperson:
ill take a look and see what I / my team can do :)

kevinburkeshyp:
maybe could try relying on your database for validation? for example with
not null constraints, or in postgres you can do CHECK positive_number
(amount > 0). pretty sure range constraints are pretty easy as well. that
way bad values can't sneak in.

On Wed, Feb 4, 2015 at 1:32 PM, Matt Apperson <notifications@github.com>
wrote:

> ill take a look and see what I / my team can do :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balderdashy/waterline/issues/824#issuecomment-72943929>
> .
>



-- 
kevin

mattapperson:
That only works on some databases, nosql solutions don't have that concept

tjwebb:
> maybe could try relying on your database for validation?

Then you have to re-implement validation for each database. This defeats the purpose of using an ORM layer in the first place.

mattapperson:
@tjwebb beyond the scope of what I feel like building (as I dont need it at all...) but perhaps we could provide a hook for adaptors to plug into to make sure validation matches?

kevinburkeshyp:
I'm confused. As a consumer of an ORM, I'm using one, maybe two database
technologies.

I'm also (probably) using multiple different languages to access a
database, which is why it's better to have these checks in the database and
not in the application, where they can be bypassed.

On Wed, Feb 4, 2015 at 2:46 PM, Travis Webb <notifications@github.com>
wrote:

> maybe could try relying on your database for validation?
>
> Then you have to re-implement validation for each database. This defeats
> the purpose of using an ORM layer in the first place.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balderdashy/waterline/issues/824#issuecomment-72956821>
> .
>



-- 
kevin

mattapperson:
@kevinburkeshyp you are assuming that all DBs implement validation. That is an incorrect assumption.

tjwebb:
[...]

Issue #658, opened 2014-09-26
Embedded documents


recidive:
As discussed in balderdashy/sails-mongo#44 and the opposite of #429. It would be nice if we could have more other than `array` and `json` to deal with embedded/nested documents in document based databases.

It was proposed by @particlebanana that we could add a `embedded` property to an association attribute like this:

```js
// Person Model
attributes: {
    name: 'string',
    age: 'integer',
    address: {
        model: 'address',
        embedded: true
    }
}

// Address Model
attributes: {
    street: 'string',
    zip: 'string',
    city: 'string',
    state: 'string'
}
```

Currently waterline doesn't support collections without adapter/connection, we have some options in how we could handle the subdocuments across different databases:

1. allow connectionless collections just to serve as schema for subdocuments models, this would make the `embedded` property obsolete, since we could assume all connectionless collections would be embedded

2. automatically set connection to the same as the parent document collection one and maybe allow to set a different connection for storing the associated document in relational databases.

For handling the embedded documents in relational databases we have few options too. Either way, I think would not require changing adapters:

1. Fallback to standard associations in relational databases (queryable)

2. Use serialized JSON for embedded documents (not queryable)

So besides the `embedded` property and what still need to be decided, in order to make this happen, we need to:

1. allow declaring embedded document associations, just like standard associations: setting `model` property will make it one-to-one and the value would be a object (embedded document), setting the `collection` property would make it one-to-many and the value would be an array of objects (array of embedded documents)

2. `create()` and `update()` methods would just validate and write the data on the main model for document based databases

What do you think?

dmarcelino:
@particlebanana, @randallmeeker, @mgenev, @devinivy what do you think of this?

leejt489:
After reading the slides on Waterline2 I've been sort of waiting to see where Waterline is heading.  But as a Mongo user I'm wondering if we can expect Waterline to support some of the key features, like embedded documents, that really improve the performance of MongoDB.  After attending one of their conferences I walked away feeling pretty sad about what I was doing in Mongo through Sails / Waterline haha.  Like I would expect the sails-mongo adapter to support embedding documents and the 'fan out on read' or 'fan out on write' patterns to keep them in sync with each other.

It's just not clear to me (maybe someone can point me to the right place) how the Sails / Waterline team is prioritizing supporting some more low level database optimizations (foreign key constraints are another come to mind) in the road map.

devinivy:
[...]

PR #1085, opened 2015-07-12
[*] fix sub attributes for mongo db

kalyuk:


dmarcelino:
Hi @kalyuk, please provide some background for this change, is there any open issue for it? If this is MongoDB specific is something that can be done in the [sails-mongo adapter](https://github.com/balderdashy/sails-mongo)?

kalyuk:
The "sails-mongo adapter" do not work because the core of "waterline" does not give an opportunity to validate nested attributes.

We can now make this type of design
older:
{
items: {
type: 'json'
}
}

now 

{
 items: [
   name: {
    type: 'string'
    minLength: 5
}
]
}

This makes it possible to use the possibility of using validation attributes nested "mongodb"
https://github.com/balderdashy/sails-mongo/issues/280

Sorry for my English :)

dmarcelino:
I understand the objective now, I think it is a similar idea to **Support embeddable association (denormalization) #428**. However this is breaking the [automated build](https://travis-ci.org/balderdashy/waterline/jobs/70578684#L9158) with many tests failing so it can't be merged as it stands now.

kalyuk:
I tried to keep backward compatibility

dmarcelino:
Cool, thanks!

glebcha:
I think this PR will resolve https://github.com/balderdashy/sails-mongo/issues/26 and https://github.com/balderdashy/waterline/issues/658 .
But all about tests now.

devinivy:
This seems to break 128 tests, many related to validation.  I wonder if this problem would be more naturally solved by #824.  You could also use `anchor` in a custom validator to achieve a similar effect.

kalyuk:
Run the tests they must pass without problems

particlebanana:
Yes, unfortunately this fails too many checks. We have embededd/denormalized schemas on the roadmap but it's not something that is planned for the immediate future.

PR #1088, opened 2015-07-13
Defaults to fix

particlebanana:
If you are using `defaultsTo` and pass in a value of `undefined` the hasOwnProperty check was failing. This fixes that and will set the value correctly.

Also will need to be hot patched on the `0.10.x` branch after merge.


dmarcelino:
Looks good to me :white_check_mark:

devinivy:
Looooooooks mergeable to me :)

PR #1114, opened 2015-08-09
Refactor normalize

dpsutton:
## Summary of some of changes

Hi. I started to learn your project by cleaning up some eslint errors. I started in the normalize.js file, which was the last on the output list, and have that file producing no new errors and warnings. There were some unknown variables I wanted to clean up but I didn't know what was going on. I ended up writing some tests and then refactoring and have the results below. There may be some controversial things about this branch so I wanted to list them here:
- Adds an eslintrc file to ~/test/.  This allows for the specification of global functions like "it", "describe". This would allow in the future differing standards for ~/lib/ and ~/test/. I saw in a discussion the project might want strict compliance in code, with warning compliance in ~/test/. This would allow for differing lint requirements to achieve this.
- In an effort to make smaller, more unit-y tests, I moved some functions out of the normalize::criteria() function. This makes the overall code more readable and has descriptive names. In an effort to make tests smaller and just test these particular code paths, I added the following to the normalize.js file:

``` javascript
    if (process.env.NODE_ENV === 'test') {
      module.exports.NotExposed = {
        StandardizedForm: StandardizeForm,
        ProcessWhereClauseAttributes: ProcessWhereClauseAttributes,
        ValidateLimitAndSkipParameters: ValidateLimitAndSkipParameters,
        NormalizeSortOptions: NormalizeSortOptions
      };
    }
```

This allows for testing inspection and calling of these functions, while not polluting namespace in production. In the test file, i expose them with simply, 

``` javascript
var units = normalize.NotExposed;
```

I realize this could be controversial, and it did require a change. The npm test runs with the environment already set to this. I had to add this to the `make coverage` options in the Makefile. A particular downside is that previously, you could run individual test files with `mocha test/unit/utils/utils.normalize.js` but they now require `NODE_ENV=test mocha test/unit/utils/utils.normalize.js`. It seems all other automated test scripts (test-unit, test-integration, coverage) now work.
- I have included comments in code trying to explain what objects look like. Should make code more readable and very abstract notions like "critieria" more tangible when reading code and deciphering function results.
- I have tried to use real objects in the test file. I figure a good place to learn what objects are under test is to use objects as close to real values as possible. The test code tries to follow a pattern of ingoing parameter and what to expect after. This is an example of the very tangible results of the functions, to help reasoning about function results easier. The function assert.deepEqual is liberally used so that object structure is more apparent to the code reader.

``` javascript
it('should recognize a sort clause', function() {
        var criteria = {where: {id: 3, sort: 'ASC'}};
        var expected = {where: {id: 3}, sort: 'ASC'};
        process(criteria);
        assert.deepEqual(expected, criteria);
      });
```
## Reasons to reject and possible changes to make compliant
- Obviously, the use of the node testing environmental variable could be objectionable to some. I could pull these functions out into a helper file and include it, and then move the tests to a new test file testing this helper. 
- Coding style. I have tried to follow styles to produce idiomatic javascript, but it is not my most used language. If there is anything that looks particularly strange, let me know. One possible objection is modifying objects in place and not returning a new copy of it. For instance, the following calls all modify the criteria object in place and never change the original pointer--so from my perspective, why return anything when it doesn't aid us. Note that StandardizeForm can change the value of the original pointer, ie, `criteria = {id: criteria}` and therefore must return. The others have a known object so they can just change attributes as necessary. Not sure which way the wind blows on this one.

``` javascript
// standardize form a little bit. If a primitive, turn into {id: primitive}
    // add where clause if not found, delete extraneous keys if no real structure to object
    criteria = StandardizeForm(criteria);


    // Move Limit, Skip, sort outside the where criteria
    ProcessWhereClauseAttributes(criteria);


    // limit and skip clauses could exist chained and now we validate they don't
    // call to skip -3 or skip -4 results. Check these arguments are above 0
    ValidateLimitAndSkipParameters(criteria);
```
- Commit style. At work, we trying to commit in small logical units. I'm not sure what etiquette is like on larger open projects like this. I could understand if you'd like all of these commits squashed into one or two at most so that the history isn't polluted with so many commits for just a simple refactor of a single file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 *.swo
 .dist
 coverage/
+npm-debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 *.swp
 *.swo
 .dist
+coverage/
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ test-integration:
 coverage:
 	@echo "\n\nRunning coverage report..."
 	rm -rf coverage
-	./node_modules/istanbul/lib/cli.js cover --report none --dir coverage/core ./node_modules/.bin/_mocha \
+	@NODE_ENV=test ./node_modules/istanbul/lib/cli.js cover --report none --dir coverage/core ./node_modules/.bin/_mocha \
 		test/integration test/structure test/support test/unit -- --recursive
 	./node_modules/istanbul/lib/cli.js cover --report none --dir coverage/adapter test/adapter/runner.js
 	./node_modules/istanbul/lib/cli.js report
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ test-unit:
 test-integration:
 	@echo "\nRunning integration tests..."
 	rm -rf node_modules/waterline-adapter-tests/node_modules/waterline;
-	ln -s $(ROOT) node_modules/waterline-adapter-tests/node_modules/waterline;
+	ln -s "$(ROOT)" node_modules/waterline-adapter-tests/node_modules/waterline;
 	@NODE_ENV=test node test/adapter/runner.js
 
 coverage:
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -187,8 +187,8 @@ Waterline.prototype.initialize = function(options, cb) {
         // Grab the schemas used on this connection
         connection._collections.forEach(function(coll) {
           var identity = coll;
-          if (hasOwnProperty(self.collections[coll].__proto__, 'tableName')) {
-            identity = self.collections[coll].__proto__.tableName;
+          if (hasOwnProperty(Object.getPrototypeOf(self.collections[coll]), 'tableName')) {
+            identity = Object.getPrototypeOf(self.collections[coll]).tableName;
           }
 
           usedSchemas[identity] = results.buildCollectionSchemas[coll];
@@ -238,8 +238,6 @@ Waterline.prototype.teardown = function teardown(cb) {
  */
 
 Waterline.prototype.bootstrap = function bootstrap(cb) {
-  var self = this;
-
   //
   // TODO:
   // Come back to this -- see https://github.com/balderdashy/waterline/issues/259
--- a/lib/waterline/utils/normalize.js
+++ b/lib/waterline/utils/normalize.js
@@ -4,13 +4,7 @@ var hop = util.object.hasOwnProperty;
 var switchback = require('switchback');
 var errorify = require('../error');
 var WLUsageError = require('../error/WLUsageError');
-if (process.env.NODE_ENV === "test" ) {
-  module.exports.NotExposed = {
-    StandardizedForm:  StandardizeForm,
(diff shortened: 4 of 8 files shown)


devinivy:
Thanks for the PR– we'll try to make our way through this in a reasonable amount of time.  I am unsure about the `NODE_ENV=test` changes, but I appreciate the issue you're trying to solve there.  The only other thing that jumped-out at me is some small style stuff.  For example, the `UpperCased` utility functions that you've created look a lot like prototyped class names.  But again, it's nice to see those blocks of code broken-out into their own routines.

@particlebanana @dmarcelino mind chiming in?

dpsutton:
yeah i've been working in C# at work and the Pascal case has creeped in on my style. The unit tests are very in style of NUnit as well.  One thing I need to do is get more consistent with the calls to assert.  I'm not sure if it was assert(expected, actual) or assert(actual, expected) but i'm sure i wasn't consistent in the code. At the very least, I think the test suites could be adapted whether the pull request is taken or not. Either way it was a good way to start learning.

dpsutton:
If you were still interested, I've updated the code to be a bit more idiomatic. I've removed the capital letter from functions, have them return values and removed the requirement to run tests with `NODE_EVN=test`. This should make the normalize feature more readable and extendable and just do some general housekeeping. The tests are expanded and offer more coverage, providing examples of what the code under test does.

particlebanana:
@dpsutton I like it, I actually committed the changes to the `.gitignore` file and `makefile` separately without seeing this one. 

My only complaint is that there are so many lines changed. It's all in the same file though so it's a bit easier to grok. Usually I would reject something like this because it doesn't add anything or fix or anything, but because normalizing the criteria is such an important pieces of the equation and something we will stick with for the foreseeable future I like that you make it easier to read and test.

I'll work my way through this eventually. Thanks!

dpsutton:
Hey thanks for looking at it. There are a few reasons why I think it might be useful.

1) It adds several more tests.

2) It adds more tests where the input/output are more explicit. When I'm trying to learn what data looks like, I like to look at the tests. But in dynamic languages, knowing what the data looks like at certain points can be difficult. Thus all of the tests I've added aren't arcane and set flags, but can be examined to see what data looks like before the function and after. Eg:

    it('should handle multiple clauses in the where clause', function() {
      var criteria = {where: {skip: 5, sort: 'ASC', limit: 7, id: 4}};
      var expected = {where: {id: 4}, skip: 5, sort: {ASC: 1}, limit: 7};
      var result = normalize.criteria(criteria);
      assert.deepEqual(expected, expected);
    });

3) The body of the function is broken into several more descriptive function calls. If you are not familiar with what normalize does, now reading the smaller body of it is more descriptive, and the tests show a more concrete example of the transformation of the data. To me, this makes the project less dense, makes reading code that you don't need the minute details of easier to get an overview, and provides simple yet concrete examples of data.

If you have any questions don't hesitate to ask. Also, if you find any non-idiomatic conventions just let me know if i can change anything. I'm C# at work so lots of things end up PascalCase instead of camelCase.

sgress454:
The normalization code has been pretty thoroughly reworked in v0.13.x, so we can't really merge this is in anymore, but we appreciate the contribution @dpsutton!

PR #1115, opened 2015-08-11
Adding default encoding option at the table level

citizenken:
Currently, there is no way to specify default encoding for a table, other than to execute a query in Sails bootstrap.js.

This change allows users to specify a model property whose value consists of an object with `charset` and/or `collation` properties. This property can then be used by a DB adapter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ var User = Waterline.Collection.extend({
   // Define an adapter to use
   adapter: 'postgresql',
 
+  // Define default encoding
+  encoding: {charset: 'utf8', collation: 'utf8_general_ci'},
+
   // Define attributes for this collection
   attributes: {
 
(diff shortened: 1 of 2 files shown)


dmarcelino:
Thanks @citizenken,

is this something that is generally used by different DBs, e.g. MongoDB, Redis? If this is something only used by very specific DBs then perhaps it's better to implement this at the adapter level?

citizenken:
@dmarcelino I'm not sure about Mongo or Redis. I know this is used by Mysql and Postgres.

I've submitted this PR https://github.com/balderdashy/sails-mysql/pull/244 for the mysql adapter to actually generate the SQL, but I could not figure out a way to get a property from my model into the adapter without adding the property to the collection object. I'm definitely open to ideas though.

particlebanana:
@citizenken part of what I have been contemplating on as far as sql migration features is a stand alone module for handling these. Adding a bunch of options to the Collection gets hard to maintain.

So for now the solution to this is to turn the `migrate: 'safe'` flag on and handle table creation using a migration helper like [Sails-Migrations](https://github.com/BlueHotDog/sails-migrations). Going forward I would like to start moving all migration logic into something that can easily be extracted out of the core framework.

Issue #1004, opened 2015-05-13
Pass the criteria into beforeUpdate


cyrilchapon:
`beforeUpdate` callback would be much more useful if it carried an `originalObject`, as third argument for example.

Retrieving original object is only achievable by either :

- Providing manually, via [policy](http://stackoverflow.com/a/28747257/4578952) or controller, the id in the request body, which is not acceptable
- [Tweaking](http://stackoverflow.com/a/29644240/4578952) and getting raw JS information on the method `this.update.arguments[0]` to get the PK, then `findOne` manually

**EDIT by @dmarcelino:** Given the interesting discussion generated here, we are re-using this thread for a feature request to add the update `criteria` to the `beforeUpdate` lifecycle in a backwards compatible way:
``` javascript
beforeUpdate: function(valuesToUpdate, cb, criteria){}
```
TL;DR: jump to https://github.com/balderdashy/waterline/issues/1004#issuecomment-101699637

See also #1122

dmarcelino:
Hi @cyrilchapon, this has the performance penalty of requiring another trip to the data store to fetch the original object. Why do you need this and why do you think it's not best to fetch the object in the controller, do what you need, and then call `.update()`?

cyrilchapon:
> this has the performance penalty of requiring another trip to the data store to fetch the original object

Then `originalObjectPK` as a third parameter, leaving the user `findOne`ing the object himself.

> why do you think it's not best to fetch the object in the controller

Because I haven't any controller, I'm using default blueprints actions

> Why do you need this

I'm trying to check if the `myobject.password` variable is a true password update, or just the hash re-sent (According to [this section](http://sailsjs.org/#!/documentation/concepts/ORM/Lifecyclecallbacks.html?q=callbacks-on-update), only "valuesToUpdate" are passed to the `update` method, but that's not always the case, when calling `.save()`. I don't have an example though, I can provide if someone needs it, but I was planning to create another issue for this)

dmarcelino:
I've never used blueprints, is it difficult/complicated to extend/override them?

>  but that's not always the case, when calling `.save()`.

You should be able to do `delete myobject.password;` when you want to avoid `beforeUpdate()` processing the password in a `.save()` call.

Have you considered creating an [instance method](https://github.com/balderdashy/waterline-docs/blob/32e3e8bed3915a043b89c42cc9f97c6a97fdf7d8/models.md#instance-methods) to deal with your case (since you want to avoid creating a controller)?

[...]

PR #1120, opened 2015-08-15
Pass valuesObject.criteria to beforeCallbacks

mmiller42:
Pass criteria to the beforeUpdate callbacks. Passes criteria as the last argument, which does not follow the Node.js convention of putting the callback last, but it would be a non-breaking change to existing implementations that expect arguments as they were.

### DIFF
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -167,12 +167,12 @@ function beforeCallbacks(valuesObject, cb) {
 
     // Run Validation with Validation LifeCycle Callbacks
     function(cb) {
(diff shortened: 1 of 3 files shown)


devinivy:
#1004 is in the 0.11 milestone, and we expect this to be a breaking change.  Want to make a PR with the breaking change to the `0.11` branch?

mmiller42:
Sure. Since you are expecting a breaking change, should I rearrange the arguments to put `criteria` before `done`, then?

mmiller42:
Opened PR #1122 with test coverage.

PR #1148, opened 2015-09-14
Switch out Object.keys for _.keys in normalize

whitfin:
This may seem like an odd change, but it aids behaviour between different Node.js versions now that there are both ES5 and ES6 flavours out there.

> In ES5, if the argument to this method is not an object (a primitive), then it will cause a TypeError. In ES6, a non-object argument will be coerced to an object.

In the current form if you pass something like `{ where: 'string' }` in Node.js 0.12.x, you will receive a `TypeError`, but in Node.js v4.x Waterline handles it according to ES6 and no `TypeError` is thrown - rather an error is generated elsewhere (haven't tracked it down) and **rejects the Promise**.

Now, the ES6 behaviour is basically the same as what `_.keys` does, so this change enforces that both ES5 and ES6 will have the same effect here. It's much more convenient for those supporting both.

FWIW I came across this with negative tests for a project; Node v4.x passed, v0.12.x failed - and the error isn't passed back to the Promise, which is what leads me to believe it should be fixed.

I should also note that someone may want to improve how this works in future, but for the moment I think it's fine to just have parity.

### DIFF
--- a/lib/waterline/utils/normalize.js
+++ b/lib/waterline/utils/normalize.js
@@ -193,7 +193,7 @@ var normalize = module.exports = {
     // false from here so the query knows to exit out.
     if (criteria.where) {
       var falsy = false;
-      Object.keys(criteria.where).forEach(function(key) {
+      _.keys(criteria.where).forEach(function(key) {
         if (Array.isArray(criteria.where[key]) && criteria.where[key].length === 0) {
           falsy = true;
         }


devinivy:
Could you give a concrete example of a call within the waterline API (i.e. use of some method on a collection or model) that causes these two different behaviors on different versions of nodejs?

iwhitfield:
@devinivy it would typically be an error case, here's one:

```
model.find({
    where: 'invalid string',
}, function (err, documents){
    // do nothing
});
```

The reason it's a concern for invalid cases like the above is that on Node 0.12.x you get an (uncaught) TypeError which crashes. On Node 4.x you get an error returned to the callback.

Here's a gist using the Express example to recreate the issue: https://gist.github.com/iwhitfield/2e89be023d6390fe6458

And example output:

```
IWhitfieldMBPR:Desktop iwhitfield$ nvm use 0.12
Now using node v0.12.7 (npm v2.13.4)
IWhitfieldMBPR:Desktop iwhitfield$ node -v
v0.12.7
IWhitfieldMBPR:Desktop iwhitfield$ node waterline.js 
/Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/utils/normalize.js:196
      Object.keys(criteria.where).forEach(function(key) {
             ^
TypeError: Object.keys called on non-object
    at Function.keys (native)
    at Object.module.exports.criteria (/Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/utils/normalize.js:196:14)
    at module.exports.find (/Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/query/finders/basic.js:220:26)
    at /Users/iwhitfield/Desktop/waterline.js:149:19
    at /Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline.js:214:7
    at /Users/iwhitfield/Desktop/node_modules/waterline/node_modules/async/lib/async.js:49:16
    at /Users/iwhitfield/Desktop/node_modules/waterline/node_modules/async/lib/async.js:263:32
    at /Users/iwhitfield/Desktop/node_modules/waterline/node_modules/async/lib/async.js:40:16
    at /Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/query/index.js:70:7
    at /Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/adapter/sync/strategies/alter.js:212:22
    at wrapper (/Users/iwhitfield/Desktop/node_modules/waterline/node_modules/lodash/index.js:3602:19)
    at applyInOriginalCtx (/Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/utils/normalize.js:421:80)
    at wrappedCallback (/Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/utils/normalize.js:320:18)
    at callback.success (/Users/iwhitfield/Desktop/node_modules/waterline/node_modules/switchback/lib/normalize.js:33:31)
    at _switch (/Users/iwhitfield/Desktop/node_modules/waterline/node_modules/switchback/lib/factory.js:48:28)
    at /Users/iwhitfield/Desktop/node_modules/waterline/lib/waterline/adapter/aggregateQueries.js:54:7

IWhitfieldMBPR:Desktop iwhitfield$ nvm use 4.0.0
Now using node v4.0.0 (npm v2.14.2)
IWhitfieldMBPR:Desktop iwhitfield$ node -v
v4.0.0
IWhitfieldMBPR:Desktop iwhitfield$ node waterline.js 
To see saved users, visit http://localhost:3000/users
null
[]
```

devinivy:
Ah, I understand the issue now!  That is a problem.  I don't think the answer to this _particular_ issue is to allow a silent failure, though.  This change allows for unexpected things to happen.  The developer certainly doesn't mean to write something equivalent to, 
```node
model.find({
    where: {
        '0': 'i',
        '1': 'n',
        '2': 'v',
        '3': 'a',
        '4': 'l',
        /* ... */
    },
}, function (err, documents){
    // do nothing
});
```

I think we should favor the ES5 behavior of throwing a hard error, since this indicates there's been some sort of developer error.  We'll have to provide an assertion before `Object.keys()` to make sure ES6 and ES5 both fail here.  It also gives us to opportunity to give the developer a useful error message about what they did incorrectly.

iwhitfield:
@devinivy yeah, I kinda agree. I'm not familiar enough with Waterline src to make that change though... how would I bubble up an error back to the callback from inside `normalize.js`?

According to Mozilla, here's a valid assertion:

```
if (typeof obj !== 'object' && (typeof obj !== 'function' || obj === null)) {
    // error
}
```

devinivy:
I'm actually advocating for a hard error that doesn't get passed through the callback (i.e. `throw Error(message)`, similar to the `TyperError` that was getting thrown), since there's no conceivable way to recover from this type of error.  The developer using waterline needs to fix something.

iwhitfield:
@devinivy I'd be against that seeing as you might be exposing your model via an API, and only hit this when someone uses it incorrectly. `TypeError` is far scarier in that case ;) It'd be safer to pass it to the callback, and let the user deal with it - typically people would log it and see that they have to fix it.

We could even just short-circuit and just pass `cb(null, [])` if this is the case, which seems to be a common thing: https://github.com/balderdashy/waterline/blob/master/lib/waterline/query/finders/basic.js#L234-L236

Alternatively, we could follow the same behaviour as the outer criteria, and if `where` is a string, do something like: https://github.com/balderdashy/waterline/blob/master/lib/waterline/utils/normalize.js#L98-L102

devinivy:
I tend to disagree with that because I think that it lets developers shoot themselves in the foot, but there is grey area here and everyone is going to feel a little differently.  My take is that the `TypeError` is scary, but it should be scary because the program was misused.

In your particular case, I would advocate for not having an HTTP API that exposes the waterline API so directly– it will probably lead to a lot of unexpected behaviors!

I think to make a decision here we'll need some more input– @tjwebb or @dmarcelino have any thoughts?  I do agree that either way, there's a change that needs to be made to unify the behavior across ES5 and ES6– thanks for bringing it up!

particlebanana:
It shouldn't crash, send out an error from the callback. Crashing is bad.

tjwebb:
Well, this adheres to my Number 1 Rule of Javascript:
1. Always use Lodash.

But it seems to possibly violate my Number 1 Rule of Programming:
1. Never fail silently


>     where: 'invalid string',

`invalid string` cannot get there through the Sails Blueprints API, which means it was put there by the programmer, right? Fallout from programmer errors is unstoppable, up to and including crashing of the app, and melting a hole in your floor. If `invalid string` made its way there from user input, then the app needs to more thoroughly validate user input. I don't actually think lodash has any role in solving either of these issues.

tjwebb:
> In the current form if you pass something like { where: 'string' } in Node.js 0.12.x, you will receive a TypeError, but in Node.js v4.x Waterline handles it according to ES6 and no TypeError is thrown - rather an error is generated elsewhere (haven't tracked it down) and rejects the Promise.

I'm not quite sure what the material difference is, here. In ES6, a thrown error that is caught would result in a rejected Promise. Do you just not have access to the original error?

zackehh:
> In ES6, a thrown error that is caught would result in a rejected Promise.

@tjwebb I'm not sure that is the case here, because I was using the Promise interface for Waterline, and the TypeError was being thrown and not given back to the Promise (which caused everything to hang, because nothing was ever resolved). I'm all for it throwing the error as long as the Promise gets it in the `catch`.

Maybe I'm wrong on this (I can't remember because it was a while ago), but I'm pretty sure this was the case. I seem to recall the only way to catch the error was with the good old `try`/`catch` (around the Promise... gross). 

I can follow up at some point to see if I'm misremembering though.

particlebanana:
@zackehh it actually looks like we are [throwing for invalid criteria](https://github.com/balderdashy/waterline/blob/master/lib/waterline/utils/normalize.js#L210-L212) so to normalize this we should do what @devinivy suggested.

Maybe change this to add the following near the top:

```javascript
if (!_.isPlainObject(criteria) || !_.isPlainObject(criteria.where)) {
  throw new WLUsageError('Invalid options/criteria :: ' + criteria);
}
```

I'm fine with keeping the `_.keys` as well. In the future we should look at returning something to the callback rather than crashing the server. But it should only be possible if you actually build an invalid query or you are sending raw user data into the `where` criteria.

sgress454:
In Waterline v0.13.x the syntax that started this whole shebang is actually valid:

```
model.find({
    where: 'invalid string',
})
```

is shorthand for `model.find({[model.primaryKey]: 'invalidString'})`.

So, this is no longer an issue!

PR #1175, opened 2015-10-02
Fixes #697

### LINKED ISSUE #697: Teardown adapter method on waterline collection is taking wrong arguments.
This piece of code that attaches non-CRUD adapter methods onto a waterline collection does not apply the correct arguments specifically for the teardown method.  Most adapter methods begin with a `connection` then a `collection`/`tableName` argument, which is what is assumed [here](https://github.com/balderdashy/waterline/blob/4653f8a18016d2bcde9a70c90dd63a7c69381935/lib/waterline/query/adapters.js#L34).

> `var args = [conn, tableName].concat(Array.prototype.slice.call(arguments));`

But the teardown method is an exception this rule.  It takes a connection, but not a table name.  What ends up happening is that you get a type error: `TypeError: string is not a function`

It's trying to call the table name string as the callback since the arguments are all out of whack.

particlebanana:
@devinivy @nethoncho this should fix the teardown connection issue.

### DIFF
--- a/lib/waterline/query/adapters.js
+++ b/lib/waterline/query/adapters.js
@@ -29,9 +29,13 @@ module.exports = function() {
 
         var tableName = self.tableName || self.identity;
 
+        // If this is the teardown method, just pass in the connection name,
+        // otherwise pass the connection and the tableName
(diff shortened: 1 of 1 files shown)


devinivy:
:+1: thanks!

PR #1190, opened 2015-10-20
Adds support for arrays on populate, like populate(['drivers','tickets'])

luislobo:
Its the usecase in between using populate as it is right now and populateAll. In this case, you can set in a single populate call all the collections and modules that need to be populated.

### DIFF
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -84,7 +84,7 @@ Deferred.prototype.populateAll = function(criteria) {
  *
  * Used for populating associations.
  *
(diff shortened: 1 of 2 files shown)


luislobo:
@particlebanana what do you think of this addition?

particlebanana:
I like it! Thanks @luislobo!

luislobo:
btw, still room for improvement here, passing also an array of conditions, that matches the array of populates... or an array of configs.  ```Invoice.find({...}).populate([{populate:'pets', criteria:{alive:true}}, {populate:'owner', criteria:{status:'active'}}])```

Issue #923, opened 2015-04-01
[Feature Request] callback afterFind


jurajpetrik:
Would it be possible to implement a  callback afterFind? E.g. I have an event with a date specified and in afterFind I could insert an attribute hasExpired..

dmarcelino:
This has also been discussed in #525 ([comment](https://github.com/balderdashy/waterline/pull/525#issuecomment-86678830)).

@aclave1, would you be interested in taking this?

I think this is a good fit for `0.11`.

cc: @randallmeeker

randallmeeker:
+1 for v0.11

anasyb:
One minute I think this should go into sails another min I think it should go into waterline. But definitely it should go somewhere!!
I was experimenting with **Model Policies** (different route policies that sails provide us with) which would also have access to the `req` as well as the returned `records`.. and run between the model action and the controller. @jurajpetrik checkout the discussion and gist here https://github.com/balderdashy/sails/issues/2760
Its a simple hook you drop into api/hooks directory... would love to hear your feedback! Gist here https://gist.github.com/anasyb/c2197a0d47ef29ee7418

anasyb:
@jurajpetrik for your particular use case have you considered an attribute function, something along the lines of:
```javascript
//in some model
attributes: {
  date:...
  //...
  hasExpired: function (){ return this.date < new Date();}
}
```

Or is it that you want it to be sent over `res` or persisted?

aclave1:
@dmarcelino I can take this. When should I do this?

dmarcelino:
How about 0.11? There is no scheduled date yet... And thanks!

jurajpetrik:
@anasyb Thanks I didn't know about attribute functions, I do want it to be sent over res though.

anasyb:
Then maybe you have two options:
option 1.  move that to the controller or the blueprint, or even a policy (depending on your case). You can keep the attr function if it has some use elsewhere.
```javascript
//api/controllers/EventController.js
module.exports = {
  findOne: function( req, res ) {
    var eventId;// = bla bla.. get it from req or from wherever you're getting it!
    Event.findOne(eventId).exec(function(err, eve){
      if ( err ) return res.serverError( err );
      if ( !eve ) return res.notFound( 'No event found with ID ' + eventId );

      var eveObj = eve.toObject();
      //temp workaround for https://github.com/balderdashy/waterline/issues/891#issuecomment-88361748
      eveObj.toJSON = undefined;
[...]

PR #1198, opened 2015-10-29
Adds afterFind callback and associated specs

britton-jb:
References issue #923

Adds afterFind callback to the `find`, `findOne`, `findOrCreate`, and `findOrCreateEach` methods, and callback runner.

Allows users to modify each record that has been found.

Includes specs for running the `afterFind` callback after `find`, `findOne`, and `findOrCreate`

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@ var User = Waterline.Collection.extend({
    * afterCreate
    * beforeDestroy
    * afterDestroy
+   * afterFind
    */
 
   beforeCreate: function(values, cb) {
@@ -545,6 +546,10 @@ Lifecycle callbacks are functions you can define to run at certain times in a qu
   - afterValidate / *fn(valuesToUpdate, cb)*
   - beforeUpdate / *fn(valuesToUpdate, cb)*
   - afterUpdate / *fn(updatedRecord, cb)*
+
+**Callbacks run on Find & FindOne**
+
+  - afterFind / *fn(values, cb)*
 
 **Callbacks run on Destroy**
 
--- a/lib/waterline/query/finders/basic.js
+++ b/lib/waterline/query/finders/basic.js
@@ -6,6 +6,7 @@ var usageError = require('../../utils/usageError');
 var utils = require('../../utils/helpers');
 var normalize = require('../../utils/normalize');
 var sorter = require('../../utils/sorter');
+var callbacks = require('../../utils/callbacksRunner');
 var Deferred = require('../deferred');
 var Joins = require('./joins');
 var Operations = require('./operations');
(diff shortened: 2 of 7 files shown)


devinivy:
Thanks for your contribution!  @particlebanana worth taking a peek.  I didn't get to read all the tests, but this essentially looks good to me.  One curiosity– `afterCreate` doesn't pass a true model instance but `afterFind` does.  I'm thinking we should fix the discrepancy by adjusting `afterCreate` rather than the work done in this PR.

britton-jb:
@devinivy I can put up a separate PR for the afterCreate model instance fix if needed.

particlebanana:
For reference this is a extension of #902

ferrants:
what's it going to take to get some movement on this PR? I think this should be rebased to incorporate recent changes. Would really like to see this get merged, let me know how I could help.

mikermcneil:
@ferrants Thanks for following up!  Would you mind putting together a [proposal](http://sailsjs.com/documentation/contributing/proposing-features-enhancements)?  You can just do it as a comment here, no need to make a separate PR or anything (I'll change the label so it gets surfaced with the others).  

Mainly I'm hoping to get an idea of the use case you're after.  @particlebanana and I have gone back and forth on this a lot of times and aside from maintenance burden / potential performance hit, there's some conceptual pros and cons to adding these.  We did make sure to set things up in WL 0.13 to where implementing arbitrary lifecycle callbacks would be straightforward to add, and consistent (i.e. you can imagine standardized lifecycle callbacks in a future release, where you always just receive a stage 2 query-- meaning you can more easily reuse logic between lifecycle callbacks).

[Here's](https://gist.github.com/mikermcneil/5fb5394c7e461afba124db4ea169d927) what we were thinking initially. (warning: slightly out of date-- we ended up keeping custom validations, and toJSON support, for example)


Basically where I'm coming from is that, if the times where this is coming up are mostly related to blueprints, then we should handle that _outside_ of waterline.  If this is more about being able to coerce something for a JSON response, then we should use toJSON instead.  If this is more about being able to coerce records more generally, when using Waterline directly in custom Node.js code, then that's more pertinent here, I'd say.

mikermcneil:
(re blueprints, see also https://gist.github.com/mikermcneil/1b87af6b6a8458254deb83a6d1cf264f)

luislobo:
I've read the proposal, I like it. First I was kind of dubious about removing the before and after validate, but since we can have before and after in other operations, then, validates can be injected there.

ChrisWorks:
Can I ask why the beforeFind and afterFind lifecycle calls are still not done 2-3 years after the they features were first requested? The use-cases has not changed; they are still very much valid.

luislobo:
Hi @ChrisWorks, Waterline has been kind of completely rewritten.
I can see from the code that there are plans to implement this (check the following links)
https://github.com/balderdashy/waterline/blob/eb86233e0304c127b91ba240fae677c8f108b8f5/lib/waterline/methods/find.js#L232
https://github.com/balderdashy/waterline/blob/eb86233e0304c127b91ba240fae677c8f108b8f5/lib/waterline/methods/find.js#L285

As with any open source project, things are done by the authors in their available time.
If you see that you can collaborate adding the support for `beforeFind` and `afterFind`, I guess the authors will be more than glad to review it. If you do so, please remember to add all the necessary unit tests too.

Best!

PR #1201, opened 2015-10-31
Set createdAt and updatedAt to the same value on creation.

ziacik:
When automatic createdAt and updatedAt property setting is enabled, while the best would be to set both times to some "transaction time", if it's not available, at least on creation the dates should be always set to the same value. The way it was implemented before, the times could differ by a millisecond (or two).


particlebanana:
Yup thanks @ziacik

PR #1225, opened 2015-11-18
Changed boolean casting method to not cast incorrect values

edupsousa:
Proposal to fix issue #1224


particlebanana:
Thanks @edupsousa

Issue #1133, opened 2015-09-02
many to many through association on through table model return array instead of single object


atiertant:
many to many through association on through table model return array instead of single object : 

      collections.user = Waterline.Collection.extend({
        identity: 'user',
        connection: 'foo',
        attributes: {
          id: {
            type: 'integer',
            primaryKey: true
          },
          cars: {
            collection: 'car',
            through: 'drive',
            via: 'car'
          }
        }
      });

      collections.drive = Waterline.Collection.extend({
        identity: 'drive',
        connection: 'foo',
        attributes: {
          id: {
            type: 'integer',
            primaryKey: true
          },
          car: {
            model: 'car'
          },
          user: {
            model: 'user'
          }
        }
      });

      collections.car = Waterline.Collection.extend({
        identity: 'car',
        connection: 'foo',
        attributes: {
          id: {
            type: 'integer',
            primaryKey: true
          },
          drivers: {
            collection: 'user',
            via: 'cars',
            dominant: true
          }
        }
      });

      Drive.findOne(1)
      .populate('car')
      .populate('user')

return

    { 
      id: 1, 
      car:[ { id: 1,createdAt: '2015-09-02T08:51:37.393Z', updatedAt: '2015-09-02T08:51:37.393Z' } ],
      user:[ { id: 1,createdAt: '2015-09-02T08:51:37.393Z', updatedAt: '2015-09-02T08:51:37.393Z' } ]
    }

istead of

    { 
      id: 1, 
      car: { id: 1,createdAt: '2015-09-02T08:51:37.393Z', updatedAt: '2015-09-02T08:51:37.393Z' } ,
      user: { id: 1,createdAt: '2015-09-02T08:51:37.393Z', updatedAt: '2015-09-02T08:51:37.393Z' }
    }

devinivy:
Anyone have time to look into this?  Failing tests are here #1134.  This will be required for proper through association support.

particlebanana:
@devinivy that's strange, if I had to guess I would say it must have something to do with Waterline Schema flagging the `Drive` collection as a join table.

atiertant:
@particlebanana one more time you where right !

weyert:
Safe to use in production?

atiertant:
@weyert in fact i think that waterline' actual manytomany through association is really buggy and should not be used in production...my PR correct some bugs...

weyert:
Thanks for the head ups. You are managing it manually now then?
[...]

PR #1134, opened 2015-09-02
Manytomanythrough fix and failling test for #1133

atiertant:
fix and failling test for https://github.com/balderdashy/waterline/issues/1133

PR #1227, opened 2015-11-20
Many-To-Many Through patch

particlebanana:
This should resolve issue #1133 discovered by @atiertant. It uses the patched version of Waterline-Schema from https://github.com/balderdashy/waterline-schema/pull/32 to use the `throughTable` flag.

If everything looks good we will need to bump the minimum version of waterline-schema once this is merged.

This is an alternative for #1134 and includes the unit test from that PR.

**EDIT:** to be clear this isn't adding any new `through` support. Only fixing an issue with the support that already existed.

### DIFF
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -126,7 +126,7 @@ Waterline.prototype.initialize = function(options, cb) {
         var junctionTables = [];
 
         Object.keys(self.schema).forEach(function(table) {
-          if (!self.schema[table].junctionTable) return;
+          if (!self.schema[table].junctionTable || self.schema[table].throughTable) return;
           junctionTables.push(Waterline.Collection.extend(self.schema[table]));
         });
 
@@ -157,6 +157,9 @@ Waterline.prototype.initialize = function(options, cb) {
         var meta = collection.meta || {};
         meta.junctionTable = hasOwnProperty(collection.waterline.schema[collection.identity], 'junctionTable') ?
           collection.waterline.schema[collection.identity].junctionTable : false;
+
+        meta.throughTable = hasOwnProperty(collection.waterline.schema[collection.identity], 'throughTable') ?
+          collection.waterline.schema[collection.identity].throughTable : false;
 
         schemas[collection.identity] = collection;
         schemas[collection.identity].definition = schema;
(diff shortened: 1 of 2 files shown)


atiertant:
@particlebanana flagging throughTable true or false is'nt suffisant to correct through associations.
the logic must be different than junctionTable because : 
junctionTable is always a junctionTable but throughTable shoud act like it on some relation only.
junctionTable contain only two relations so populate skip parent relation and chose the other to find the child one but a throughTable contain more associations...so this can't be done like this.

i propose to add this in schema to correct the problem : 

    throughTable : [
       "driver.cars": "car",
       "car.drivers": "driver"
    ]

here we save the other part of throughTable in a relation dot notation key.
like this we can check if we must consider it as a  throughTable or simply as a collection depending the context and know the correct association of the throughTable use in populate.

particlebanana:
@atiertant this fixes the issue shown in the test. It only loads the `throughTable` once with the original definition which allows you to use the `Drive` table as a normal collection or as a `junctionTable`.

I'm not sure I understand the issue. Using `via` should map up everything correctly for other associations on the `Drive` collection. Could you put together a gist showing something that doesn't currently work (after this patch and the schema patch)? 

If you are saying you want three levels of results for many-to-many through that is a separate feature. I think most of that logic would live in waterline-cursor but would take some work. That is separate from this issue.

So to clear up this allows you to have a collection that can work as both a normal collection and a junction table.

To have a collection that returns the values on the throughTable when populating either side (User, Car) is a different issue and involves work with the cursor and elsewhere.

atiertant:
@particlebanana i was not talking about a new feature but bug fix.(of course using dot notation open deep features but this is not the question here)
i will pr a new set of tests to show you what i was talking about next week.

atiertant:
@particlebanana i had some tests,please a look at this: https://github.com/atiertant/waterline/blob/manytomanythrough/test/unit/query/associations/manyToManyThrough.js

particlebanana:
@atiertant thanks, looking into it.

particlebanana:
@atiertant Ah I see now why this is needed. It was dumbly just looking for the other key in the junction table. Ok I will close mine and go with yours.

The `.add()` and `.remove()` test will fail in your test case but I have a fix for that I will PR after these are merged. Thanks for all your help and sorry it took so long for me to wrap my head around.

PR #1237, opened 2015-11-30
Returning Error instead of a string

GeorgeSapkin:


devinivy:
Wow, surprised to see this!  Looks good to me.

particlebanana:
Thanks @GeorgeSapkin!

PR #1239, opened 2015-11-30
Removed broken FoundationDB SQL Layer link

GeorgeSapkin:
FoundationDB doesn't exist as a separate company or a product anymore.


particlebanana:
Thanks! You're on :fire:!

Issue #763, opened 2014-12-18
Is it possible to create a record without assigning a PK & saving?


cha0s:
I'd like to be able to create an instance of a model without actually saving it or generating a PK (ID would equal null). Is this possible? I'd like to get an instance with all defaults populated, etc. but not actually save it.

particlebanana:
This *should* work: `new Pet._model(values);`

cha0s:
Using an underscored method is a code smell. Can we have this as a public API, please?

particlebanana:
Sure mind putting together a pull request?

cha0s:
Yup I'll get to this later tonight


Sent from my T-Mobile 4G LTE Device

<div>-------- Original message --------</div><div>From: Cody Stoltman <notifications@github.com> </div><div>Date:01/16/2015  12:17 PM  (GMT-06:00) </div><div>To: balderdashy/waterline <waterline@noreply.github.com> </div><div>Cc: Ruben Rodriguez <cha0s@therealcha0s.net> </div><div>Subject: Re: [waterline] Is it possible to create a record without assigning a
  PK & saving? (#763) </div><div>
</div>Sure mind putting together a pull request?

cha0s:
Do you have a preference about the syntax? In the project I desired this functionality in, I augmented all collections with `.instantiate`, so I could do `var user = User.instantiate();`.

Does that seem like a god name? Should we go with `instance`, or something else do you think?

Might as well talk about it a little before I bother with a PR :)

andrewjmead:
+1 for this. I wouldn't mind helping out with a PR as waterline has saved me a ton of development and maintenance time.

My use case is creating multiple instances of a model, but only if they are all valid.

rishabhmhjn:
Was any method to support this feature added?

devinivy:
No, there wasn't, but I would love to see a PR for this!  Using `Collection._model({/* record */})` still works (and I _expect_ it will work until there is a breaking change in waterline), though it doesn't feel great to use the private method.

ejlangev:
`Collection._model({ /* data */})` doesn't seem to work very well.  I'm using waterline 0.10.26.  The method initially works and returns an instance of the model but if I try to call save on that model I immediately get an error "Error: No Primary Key set to update the record with! Primary Key must have a value, it can't be an optional value."

Some example code:

```js
var user = User._model({ email: 'foo@bar.com' });

user.save()
  .then(function() {
    console.log(user.id);
  })
[...]

PR #1262, opened 2015-12-23
Allow calling save on an unpersisted record

ejlangev:
First pass on an implementation of https://github.com/balderdashy/waterline/issues/763.  Seems pretty simple but might be totally wrong.  Let me know if it is so I can correct.  @devinivy

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -74,15 +74,30 @@ module.exports = function(context, proto, options, cb) {
       }, next);
     },
 
-    // Update The Current Record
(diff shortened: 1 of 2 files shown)


ejlangev:
Also this is probably against the incorrect branch.  If so let me know what branch it should be against so I can fix.

particlebanana:
Hey @ejlangev can you spec out in a comment exactly what the API looks for using this?

ejlangev:
```js
var record = new MyModel({ a: 1, b: 'a' });
return record.save()
  .then(function() {
    // Do something after the save if necessary
  })
```

shvelo:
Does this add support for `new MyModel()`?

PR #1266, opened 2015-12-26
Reorganize Roadmap

randallmeeker:
This reorganized the road map. The hope here is that we speak more truthful to Waterline development. Features that do not have an active developer on them should be left in backlog till someone picks up the torch.


particlebanana:
Thanks @randallmeeker

PR #1268, opened 2015-12-28
Through association id datetime fix

elennaro:
Resolves Issue #1265.

I'll do and fix anything you tell! Just please - merge this.

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/toObject.js
+++ b/lib/waterline/model/lib/defaultMethods/toObject.js
@@ -102,8 +102,13 @@ toObject.prototype.addAssociations = function() {
     // Handle belongsTo attributes
     var record = this.proto[association];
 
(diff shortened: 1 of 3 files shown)


particlebanana:
@elennaro I'll look into this PR as it shouldn't crash the server but you shouldn't use datetime as a primary key. It's not guaranteed to be unique.

particlebanana:
Actually I should have looked at this more. `item` is missing the `var` statement and I'm not sure what the end goal is as just assigning it to item doesn't do anything. Going to revert this for now.

elennaro:
@particlebanana 
Oh gee! Sorry! This was a stupid way to fix the issue, and I forgot about assigning.
I'm fixing this right now.
I'll make a new PR asap.

P.S. Thanks, for responding so quickly, we are suffering from lack of 'datetime' much.

PR #1269, opened 2015-12-29
Through association id datetime (try 2).

elennaro:
Second try to resolve Issue  #1265.

I've refactored the part like this:

```
...
if (record instanceof Date) {

      var item = new Date(record);
      _.extend(item.__proto__ , record.__proto__);

      this.object[association] = item;

      // _.isObject() does not match null, so we're good here.
    }
...
```

The question is do we need `_.extend(item.__proto__ , record.__proto__);` or it has no sense here, since no one extends Date with custom methods?

@particlebanana 
I'll fix and rewrite anything you say.

Thanks for your patience and care.
I never committed to such a big and serious enterprise projects, so I feel a little nervous :)

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/toObject.js
+++ b/lib/waterline/model/lib/defaultMethods/toObject.js
@@ -1,4 +1,3 @@
-
 /**
  * Module dependencies
  */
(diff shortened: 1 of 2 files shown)


particlebanana:
@elennaro hey don't be worried! We'll get it figured out and merged in. I'll try and take a look at this later today. Thanks for your work!

elennaro:
@particlebanana thanks a lot!
You guys are wonderful!

elennaro:
I so sorry to bother you. 
But we need some fix for Issue #1265 to be able to update from 0.10.27.

If you don't like something in this PR, I will rewrite it as you say.

particlebanana:
Ok @elennaro this one looks good. I think using dates as primary keys is a horrible solution but making sure it doesn't crash is more important. Thanks for the work, I'll get a patch published shortly.

elennaro:
@particlebanana thank you vary much! You saved our day!
You're totally right about date as PK :)

Issue #745, opened 2014-11-28
Error on init: TypeError: Cannot read property 'identity' of undefined


marcelklehr:
I get this error when calling waterline.initialize()
```
node_modules/waterline/lib/waterline/adapter/sync/strategies/alter.js:42
 = _.find(self.query.waterline.schema, {tableName: self.collection}).identity;

TypeError: Cannot read property 'identity' of undefined
    at afterDescribe (node_modules/waterline/lib/waterline/adapter/sync/strategies/alter.js:42:91)
    at bound (node_modules/waterline/node_modules/lodash/dist/lodash.js:957:21)
    at applyInOriginalCtx (node_modules/waterline/lib/waterline/utils/normalize.js:416:80)
    at wrappedCallback (node_modules/waterline/lib/waterline/utils/normalize.js:315:18)
    at _normalizeCallback.callback.success (node_modules/waterline/node_modules/node-switchback/lib/normalize.js:33:31)
    at _switch (node_modules/waterline/node_modules/node-switchback/lib/factory.js:35:28)
    at node_modules/sails-memory/lib/database.js:163:5
    at null._onTimeout (node_modules/sails-memory/lib/database.js:122:5)
    at Timer.listOnTimeout (timers.js:133:15)
```

andrewjmead:
I'm getting that same error. I was able to debug it to be an issue with capitalization in the identity name. It would run once fine, then it would error out the second time. I updated the name to lower case and blasted the DB away and everything was fine.

```
// Name that did not work
identity: "linkStatus",

// Name that did work
identity: "linkstatus",

```

emersion:
@andrewjmead Same problem, same solution. Thanks!

andrewjmead:
@emersion Thanks for following up!

jeremiahrhall:
The upper to lower case change worked for me as well.

tjwebb:
Yea, by the way, when you don't manually set `identity`, sails automatically sets it to be lowercase.

devinivy:
@tjwebb I thought identities were always lowercase–how do you "manually" set an identity with case?  If identities are always lowercase and that's not in the docs, let's add it.  Also, if identities should be lowercase, we should add a useful error message.

tjwebb:
I don't see it documented anywhere but other github comments. Here's some relevant code: https://github.com/balderdashy/waterline/blob/master/lib/waterline.js#L116

[...]

PR #1278, opened 2016-01-18
added to roadmap: do not mess with identity case (see #745)

japel:
see #745


particlebanana:
Thanks @japel. For anyone coming across this in the future we have a new [Contribution Guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) which specs out what requests like this should like when being made.

Issue #1091, opened 2015-07-16
save() Callback Returning Record Inconsistent with Database


itsalaidbacklife:
This issue appears similar to [975](https://github.com/balderdashy/waterline/issues/975) as well as [2073,](https://github.com/balderdashy/sails/issues/2073) but it is different in that while the save() call fails to persist the local data to the database, waterline throws no error. Essentially, when making several nested find()'s, then calling add() and remove() on different collections in the returned records and then nesting the save() calls, the local copies of the records returned by save() are **correct,** but the data is **not** updated on the database. The records returned by save() are inconsistent with the records found at localhost:1337/modelName. Is this a bug, or am I misusing save()?

Here are the relevant models:

 In Card.js:

    attributes: {
      rank: {type: 'integer'},
      suit: {type: 'integer'},
      deck: {model: 'game'},
      scrap: {model: 'game'},
      hand: {model: 'player'},
      points: {model: 'player'},
      runes: {model: 'player'},
      stackedTwo: {model: 'game'},
      targetId: {type: 'integer'},
      attached: {model: 'card'},
      attachments: {collection: 'card', via: 'attached'},

      class: {type: 'string', defaultsTo: 'card',}
    }

In Player.js:
       
    attributes: {
      socketId: {type: 'string'},
      currentGame: {model: 'game'},
      pNum: {type: 'integer'},
      hand: {collection: 'card', via: 'hand'},
      points: {collection: 'card', via: 'points'},
      runes: {collection: 'card', via: 'runes'},
      frozenId: {type: 'integer', defaultsTo: null},
      tempString: {type: 'string', defaultsTo: ''},
    }

Below is the action in our controller:
It is supposed to find two players, one of which is the 'thief' and one of which is the 'victim'. The 'theif' player removes a jack from his hand collection and adds that jack to the 'attachments' collection of a card in the 'victim' player's runes collection. The chosen card is then removed from the victim's runes and added to the 'thief' player's runes. Both players are then saved in a nested fashion, so that the callback passed to the innermost save() has access to all of the fully updated records, for the purpose of running a publishUpdate() that we do not currently call. the lines `console.log(savedP0)` and `console.log(savedP1)` print the exact expected data, but checking the database at localhost:1337/player reveals that the only change made on the database was removing the jack card from the 'thief' player's hand. None of the add() calls, nor the other remove() calls, besides the one on the 'thief' player's hand, persist on the database.

	jackBug: function (req, res) {
[...]

Issue #1037, opened 2015-05-30
.save() and .populate() aren't friends


anasyb:
```
User.findOne('xyz').populate('posts').exec(function(err, user){
  user.comments.add('someCommentID');
  user.save();
})
```
This is trying to re-insert all posts into user?! Of course it throws off an `already exists` error (and sometimes even goes ahead and inserts duplicate records into the join table https://github.com/balderdashy/sails-mongo/issues/288 !!!). Somehow related to https://github.com/balderdashy/waterline/issues/975 and the other issues referenced in there, but not sure its the same thing.
On mongo.

P.S.: not sure if `user.comments.add('someCommentID');` is needed to reproduce the error, but that's what I have in my code. Maybe its just the `.save()`.

devinivy:
Related: this old PR #898 fixing #858, where `save` and `populate` were not behaving very nicely together.  In your example, are Users and Posts in a one-to-many or many-to-many relationship?

anasyb:
Thanks @devinivy, as awesome as always!
- sorry haven't caught that issue and PR, looks very similar
- Yeah I think I am still on 10.18.
- This happened in a one-way relationship: user has many posts, but post has no user. I think (but not sure) waterline assumes it's many-to-many.

devinivy:
I highly suggest upgrading past 0.10.18.  There's a data integrity danger if you populate an incomplete collection (say, by using population criteria) then save.

dmarcelino:
Hey @anasyb, did updating waterline fix the issue?

dmarcelino:
Closing this as inactive, please re-open if the problem persists.

PR #1295, opened 2016-02-03
Remove second argument to .save()

particlebanana:
> This is a breaking change and will require us to go ahead and push out a smaller version of `0.11`.

This may be controversial but I think it's the right thing to do. The current behavior of `.save()` is awful as documented in #441 #1091 #1037 #954 and probably more. Right now it populates all the attributes which can easily overflow memory. There is a flag to mitigate some of this but it shouldn't happen in the first place.

Trying to determine what to populate based on what attributes are currently set in the model sounds nasty as well. Removing this attribute makes the API cleaner and more predictable as well as a tad bit faster because it has do at minimum one less query.

I want to continue removing pieces related to model serialization in future releases as well so nested/graph operations are probably next on the chopping block.

So to summarize this changes the function signature of the callback passed to `.save()` from `.save(function(err, updatedRecord) {})` to `.save(function(err) {})`.

### DIFF
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -191,6 +191,16 @@ module.exports = function(context, proto, options, cb) {
       return cb(error);
     }
 
+    // Reset the model attribute values with the new values.
+    // This is needed because you could have a lifecycle callback that has
+    // changed the data since last time you accessed it.
+    // Attach attributes to the model instance
+    var newData = results.updateRecord[0];
+    _.each(newData, function(val, key) {
+      proto[key] = val;
+    });
+
+    // If a promise, resolve it
     if (deferred) {
       deferred.resolve();
     }
--- a/lib/waterline/model/lib/defaultMethods/save.js
+++ b/lib/waterline/model/lib/defaultMethods/save.js
@@ -7,9 +7,6 @@ var removeAssociation = require('../associationMethods/remove');
 var hop = require('../../../utils/helpers').object.hasOwnProperty;
 var defer = require('../../../utils/defer');
 var noop = function() {};
-var optDefaults = {
-  populate: true
-};
 
 /**
  * Model.save()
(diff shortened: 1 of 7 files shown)


devinivy:
Does this affect `save()`'s promised interface in the way one would expect?

particlebanana:
@devinivy it changes the promise resolver from:

```javascript
if (deferred) {		
  deferred.resolve(data);		
}
```

```javascript
if (deferred) {		
  deferred.resolve();		
}
```

I don't use promises does this cause some unexpected issue?

devinivy:
That seems right!  Just makin' sure, since it seems like the type of thing that could get buggered.  Still thinking about two things,

1. Can we reset the values on the saved instance with the new values from the db?  Imagine saving an object, defaults being applied to an attribute, the update occurring successfully, then the model instance not reflecting the defaults that were applied.
2. Can we make an additional breaking change to pass a legit `Error` rather than an array of objects representing failed transactions?

particlebanana:
@devinivy I think both of those requests are valid. I'd like to try and get this out today. What should the `Error` look like you think?

devinivy:
Cool!  For the error maybe something simple like,
```js
var err = new WLError('Some associations could not be added or destroyed during save().');
err.failedTransactions = failedTransactions; // The array of failed transactions
```

pauln:
Is there any reason that the population options were removed completely, rather than simply changing the default (to `false`, so that nothing is populated by default but it's possible to specify keys to populate - or set it to `true` if you really want to populate all associations)?  It doesn't seem to be possible to `populate()` after a `save()` - I get `myObj.save(...).populate is not a function` - so with the population options gone, I can't see a way to populate the associations at all.  This means that although I had an association populated before saving, once I've saved, it's been flattened with no way to populate it again - which seems like a strange side-effect of saving.

If I'm missing something, please enlighten me - but otherwise, can we please either have the population options back (with appropriate defaults) or get support for `.save().populate()` implemented?

PR #1306, opened 2016-02-19
Meta attribute property

particlebanana:
Part of the confusion and discussion about migrations has been that the attributes define properties in the physical storage layer. It's not ideal to continue adding adapter specific features to attribute definitions. This adds a reserved property to attribute definitions labeled `meta` that allows anything to be passed down to the adapter's `registerConnection` method.

The goal being that any adapter specific details can be defined and passed through. At the same time it cleans up the arguments being passed to `registerConnection` so that instead of trying to make sense of all the internal Waterline details you simple get a dictionary with the keys being the collection identity and each collection containing the following data:

* `definition` - the normalized table definition that was already being used by adapters. This stays the same.
* `attributes` - user defined attributes, including the `meta` field mentioned above.
* `meta` - a pre-existing object that defined extra generated metadata about a collection, such as the fact that it is a junctionTable.

I would still like to move all migrations out of Waterline in future releases but this will continue to be part of the spec.

```javascript
attributes: {
    email: {
      type: 'string',
      email: true,
      meta: {
        foo: 'bar',
        bar: function(hello) {
          return hello;
        }
      }
    }
  }
```

### DIFF
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -84,7 +84,7 @@ Waterline.prototype.initialize = function(options, cb) {
   // Allow collections to be passed in to the initialize method
   if (options.collections) {
     for (var collection in options.collections) {
-      this.loadCollection(_.cloneDeep(options.collections[collection]));
+      this.loadCollection(options.collections[collection]);
     }
 
     // Remove collections from the options after they have been loaded
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -160,6 +160,7 @@ Waterline.prototype.initialize = function(options, cb) {
 
         schemas[collection.identity] = collection;
         schemas[collection.identity].definition = schema;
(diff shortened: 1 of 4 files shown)


mikermcneil:
Ooh man, so glad this still merges after what I just did.  Very nice.

@particlebanana heads up- the callback from `Validator.prototype.validate()` now (finally) takes a first argument: `err`.  Originally, async custom validations were added via a PR, which was no big deal.  But since this was back before we were crotchety and jaded, it ended up getting moved around a bunch, and there was a throwing error that would crash the process.  That's taken care of, relevant tests are updated, and luckily it is encapsulated by an interface that didn't require changing too many places in the code base.

Anyway, I was mainly just checking to be sure I didn't mess this PR up.  And it looks like we're good.

PR #1307, opened 2016-02-19
Criterias with extra options can pass through.

gaurav21r:
This is a really *tiny* change and i don't think it would *disrupt* any tests etc.

Basically, our team works on custom Waterline Adapters and one of things that is required is to have changes to the Query language. Currently if we pass custom options to the criteria, it is automatically `mixin`ed into `where`.

So `model.find({opt1: {})` becomes `model.find({where: {opt1: {}})`.

If we explicitly have a `where` then this doesn't happen so
`model.find({where: {name: 'abc'}, opt1: {}})` remains unchanged.

**BUT** :exclamation: `model.find({where: {}, opt1: {}})` :arrow_right: `model.find({where: {opt1: {}}})`

The only thing this PR is doing is `model.find({where: {}, opt1: {}})` remains unchanged.
I think this is a reasonable intended behavior!

### DIFF
--- a/lib/waterline/utils/normalize.js
+++ b/lib/waterline/utils/normalize.js
@@ -183,11 +183,6 @@ var normalize = module.exports = {
       delete criteria.where.select;
     }
 
-    // If WHERE is {}, always change it back to null
-    if (criteria.where && _.keys(criteria.where).length === 0) {
-      criteria.where = null;
-    }
-
     // If an IN was specified in the top level query and is an empty array, we can return an
     // empty object without running the query because nothing will match anyway. Let's return
     // false from here so the query knows to exit out.


particlebanana:
@gaurav21r I was actually mocking something similar to this up today. Basically we need a way to pass arbitrary data down from a query to the adapter that isn't processed by Waterline. This could include things such as an active connection to a database or extra config options, etc. 

I thought of adding it to the criteria object as a quick fix but it seems a bit hacky and would be harder to ensure it isn't modified along the way. I would like a standalone argument that gets passed in.

The current function signatures are pretty widespread so I'm hesitant to break those. Would everyone be completely disgusted if I added an argument **AFTER** the `cb` argument?

```javascript
find: function find(connectionName, collectionName, options, cb, metadata) {}
```

gaurav21r:
@particlebanana Sorry for the delayed response. Great to know we are not alone with this use case!
Our requirement (a little lopsided though) is to get User credentials from a Database in order to communicate via a REST api.

Your proposal seems superior. But I think it breaks the **node style callbacks** which is a term oft used in the JS world (dunno what it means exactly). `cb` is no longer the **last** argument to the function. Will it break things like [Promisify](http://bluebirdjs.com/docs/api/promise.promisify.html) ? 

@particlebanana Why I chose this format was because, when the `where` is already a non null object, it works as expected. `model.find({where: {name: 'abc'}, opt1: {}})` remains unchanged. But why does `model.find({where: {}, opt1: {}})` :arrow_right: `model.find({where: {opt1: {}}})` ?

This seemed non standard behaviour and so I thought this was the best way to handle this.

devinivy:
I wouldn't love an argument after the callback.  I'd much prefer a `metadata` key in the options.

mikermcneil:
> I wouldn't love an argument after the callback. I'd much prefer a metadata key in the options.

> it breaks the node style callbacks

@gaurav21r @devinivy completely agree with you guys-- it's just that we clone/normalize those options a great deal, and existing adapter code may be relying on having exactly the expected keys in there.  The argument after the callback is sort of an immediate-term solution that we can use without breaking existing adapters.  Longer term, I think it would be great to add an `opts` key (like we're doing at the driver level), and allow it to be nested inside of populates as well.  The important thing is that this extra metadata would need to be insulated from any kind of cloning/JSON-serialization, and would also need to not be recursively parsed (since it could be any type, and might contain circular references).

particlebanana:
@gaurav21r @mikermcneil this would be a temporary measure to add the functionality in without breaking all of the community adapters out there. We don't even have to document it. Then when we are ready to push out the breaking change we tag it `1.0.0` and move to the normal style function signatures.

We already wrap the model functions with a promise interface using the deferred library so you can use `.then` and `.catch`. Is there a situation where you would still need to use Promisify?

Having just added the projections support there are a lot of things touching the criteria and tacking on things that will be passed through is a pretty big hurdle.

particlebanana:
Ok the extra argument is added in #1325 and will be published in `0.12.0`.

PR #1310, opened 2016-02-19
Projections

particlebanana:
### DO NOT MERGE

This PR represents the initial work for projections on both the parent and child criteria. It finally brings official support for `select` and adds a `.select()` method to the deferred object. It also adds support for using `select` inside of criteria in a `populate`. 

One thing to note is that any primary keys and foreign keys used in a join will always be selected. This is unavoidable due to the way we currently build up join queries and combine results in cross-adapter populates.

The first thing that needs to happen is tests in [Waterline Adapter Tests](https://github.com/balderdashy/waterline-adapter-tests) to ensure things won't break across the officially supported adapters. There is a corresponding PR in [Waterline Sequel](https://github.com/balderdashy/waterline-sequel/pull/80) that needs to be brought in for the SQL adapters and I'm sure there will be some slight tweaks that need to happen in the Mongo adapter.

**Example Use:**

```javascript
// Deferred Object Style
User.find()
.select(['name', 'age'])
.where({
  age: {
    '>': 21
  }
})
.exec(function(err, users) {});
```

```javascript
// Populate projections
User.find()
.select(['name', 'age'])
.populate('pets', { select: ['breed', 'name'] })
.exec(function(err, users) {});
```

```javascript
// Callback style
User.find({ select: ['name', 'age'] }, function(err, users) {});
```

### DIFF
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -189,9 +189,17 @@ Deferred.prototype.populate = function(keyName, criteria) {
       select.push(val.columnName);
     });
 
-    // Ensure the PK and FK on the parent are always selected - otherwise things
+    // Ensure the PK and FK on the child are always selected - otherwise things
     // like the integrator won't work correctly
     if(join.collection) {
+      var childPk;
+      _.each(this._context.waterline.schema[attr.references].attributes, function(val, key) {
+        if(_.has(val, 'primaryKey') && val.primaryKey) {
+          childPk = val.columnName || key;
+        }
+      });
+
+      select.push(childPk);
       select.push(attr.on);
     }
 
@@ -221,7 +229,7 @@ Deferred.prototype.populate = function(keyName, criteria) {
         if(_.has(val, 'collection')) {
           return;
         }
-        
+
         // Check if the user has defined a custom select and if so normalize it
         if(customSelect && !_.includes(criteria.select, key)) {
           return;
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -191,15 +191,17 @@ Deferred.prototype.populate = function(keyName, criteria) {
 
     // Ensure the PK and FK on the child are always selected - otherwise things
     // like the integrator won't work correctly
-    if(join.collection) {
-      var childPk;
-      _.each(this._context.waterline.schema[attr.references].attributes, function(val, key) {
-        if(_.has(val, 'primaryKey') && val.primaryKey) {
-          childPk = val.columnName || key;
-        }
-      });
+    var childPk;
+    _.each(this._context.waterline.schema[attr.references].attributes, function(val, key) {
+      if(_.has(val, 'primaryKey') && val.primaryKey) {
+        childPk = val.columnName || key;
+      }
+    });
 
-      select.push(childPk);
+    select.push(childPk);
+
+    // Add the foreign key for collections
+    if(join.collection) {
       select.push(attr.on);
     }
 
--- a/lib/waterline/query/deferred.js
+++ b/lib/waterline/query/deferred.js
@@ -511,9 +511,8 @@ Deferred.prototype.set = function(values) {
  */
 
 Deferred.prototype.exec = function(cb) {
-
   if (!cb) {
-    console.log(new Error('Error: No Callback supplied, you must define a callback.').message);
+    console.log('Error: No Callback supplied, you must define a callback.');
     return;
   }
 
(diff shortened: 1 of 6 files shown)


lib/waterline/query/deferred.js
```diff
@@ -271,6 +273,22 @@ Deferred.prototype.populate = function(keyName, criteria) {
...
+ * @param {Array} attributes to select
+ * @return this
+ */
+
+Deferred.prototype.select = function(attributes) {
+  if(!_.isArray(attributes)) {
+    attributes = [attributes];
+  }
+
+  this._criteria.select = attributes;
```
shamasis: It should append to existing select and not replace it. Will allow to call select twice.

shamasis:
For the populates, especially joins - WL-schema also needs to honour the select arrays. FYI. :-)

lib/waterline/query/deferred.js
shamasis: I mean, what do you think?

particlebanana: Ah good call!

particlebanana:
@shamasis I have another PR for [wl-sequel](https://github.com/balderdashy/waterline-sequel/pull/80) if that's what you meant. If not what does wl-schema need for this?

nicesu:
hello , now  i can use .selet()  and  .populate(''，{select: []}) ？

cjy37:
+1

mikermcneil:
> One thing to note is that any primary keys and foreign keys used in a join will always be selected. This is unavoidable due to the way we currently build up join queries and combine results in cross-adapter populates.

:+1: 



@particlebanana looks great man.

To get reaaallllll specific for a second, what happens if I:

1. `.select('*')`   / `.populate('direwolves', { select: '*' })`
2. `.select(['*'])`  / `.populate('direwolves', { select: ['*'] })`
3. `.select(['*', 'coatColor'])`   / `.populate('direwolves', { select: ['*', 'coatColor'] })`
4. `.select([])`    / `.populate('direwolves', { select: [] })`

My thoughts are:
1. not valid-- call cb with invalid criteria error  (if we're feeling fancy later we can expand to `['*']` automatically once this is well-tested)
2. not sure on this, because of the way it could affect driver-level stuff.  I think we need to figure it out there first maybe.
3. depends on 2 (but my intuition is to call cb with error)
4. interpreted as _only_ fetching the primary key and any necessary foreign keys.

particlebanana:
@mikermcneil started building out tests for projections in the adapter test suite and added the following checks.

https://github.com/balderdashy/waterline-adapter-tests/pull/106

`.select(['*'])`  / `.populate('direwolves', { select: ['*'] })`

Returns all fields for the record. Same as `select * from 'foo'`

---------------------------------

`.select('*')`   / `.populate('direwolves', { select: '*' })`

Gets normalized into `select: ['*']` and returns all fields. This is done in the [criteria normalize](https://github.com/balderdashy/waterline/blob/projections/lib/waterline/utils/normalize.js#L187-L195) function.

---------------------------------

`.select(['*', 'coatColor'])`   / `.populate('direwolves', { select: ['*', 'coatColor'] })`

Gets normalized into `select: ['*']` and returns all fields. If a star is used at all you will get all fields.

---------------------------------

`.select([])`    / `.populate('direwolves', { select: [] })`

Only returns the primary key for the record.

Salakar:
Is this actually removing the unnecessary fields on waterline itself by modifying the returned data or is this doing projection queries on the db side (i.e mongo)?

particlebanana:
@Salakar this supports adapter level projections so it's up to the adapter to build them correctly. The latest version of Waterline adds this support and it's implemented in the sql adapters as well as mongo I believe.

luislobo:
I have an other case: if `.select(false)` or  `.select(undefined)` it should just be treated as `'*'`, that is, ignore the `select`. I have a caching function that could benefit of a select statement: `CacheService.find(Model, criteria, populate, populateCriteria, select, sort)`. I could call it like: `CacheService.find(User, {email:'luislobo@gmail.com'})`, or `CacheService.find(User, {email:'luislobo@gmail.com'},'groups',false,['email','name','groups'])`

What do you think? From the code, I see that undefined or false is not handled. https://github.com/balderdashy/waterline/blob/master/lib/waterline/utils/normalize.js#L193

May be it's handled somewhere else?

particlebanana:
@luislobo we should add some [tests](https://github.com/balderdashy/waterline-adapter-tests/blob/master/interfaces/queryable/projections.test.js) for those cases first. But that sounds good.

luislobo:
I can create a PR for this

TahaBoulehmi:
What happened to this feature? Is it supported? It doesn't work on Sails 1.2.3.

TahaBoulehmi:
I made a pull request to solve the one-to-one relationship issue:
https://github.com/balderdashy/waterline/pull/1613

PR #1316, opened 2016-02-24
Type should only be forced to integer if autoIncrement is set to true

janhapke:
For now this is more of an illustration for #1314 - I haven't yet had time to figure out how to run the tests for waterline. Might do that later.

### DIFF
--- a/lib/waterline/core/schema.js
+++ b/lib/waterline/core/schema.js
@@ -125,7 +125,6 @@ Schema.prototype.objectAttribute = function(attrName, value) {
       // Set schema[attribute].autoIncrement
       case 'autoIncrement':
         attr.autoIncrement = value[key];
-        attr.type = 'integer';
         break;
 
       // Set schema[attribute].unique
@@ -171,6 +170,11 @@ Schema.prototype.objectAttribute = function(attrName, value) {
         attr.alias = attrName;
         break;
     }
+  }
+
+  // auto increment only works for integer
+  if (value['autoIncrement'] == true) {
+      attr.type = 'integer';
   }
 
   return attr;
--- a/test/unit/core/core.schema/schema.autoValues.js
+++ b/test/unit/core/core.schema/schema.autoValues.js
@@ -82,6 +82,47 @@ describe('Core Schema', function() {
     });
   });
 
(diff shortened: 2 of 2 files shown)


devinivy:
Seems good to me.  Honestly, I don't think that auto-increment should force the type to integer at all, but I'm not sure who that would affect.  "Auto-increment" should possibly be broadened to, for example, indicate that UUIDs should be used in DBs that do not support actual integer incrementing.  Or other strategies for generating ids.

mikermcneil:
@janhapke Nice, thanks!  I think this also solves https://github.com/balderdashy/sails/issues/3791

mikermcneil:
@devinivy btw, a few other thoughts re: future stuff w/ autoPK:

Right now Waterline mostly cares about auto-increment only inasmuch as it affects auto-migrations.  But going forward, the plan is to externalize auto-migrations support (e.g. move it to sails-hook-orm).  So I'm not really sure Waterline will even need to know about whether incrementing is happening or not...  Well, except that Waterline _will_ still need to know that a particular attribute is going to be automatically set by the db on a `.create()`.  But even then, only when it's the primary key, yeah?  Kind of like a special, dynamic `defaultsTo`, but automatically provided by the database.

So at that point-- what if, for v1.0, we just removed [autoIncrement](http://sailsjs.org/documentation/concepts/models-and-orm/attributes#?autoincrement) from Waterline, and instead added a new flag for attribute defs (e.g. `automatic: true` or `autoDefaultsTo: true`).  That flag would have _no impact_ on `type`, but it would not be allowed to coexist with `required: true` or `defaultsTo`.  Its sole purpose would be to support use cases like https://github.com/balderdashy/waterline/blob/3366d35408732fdde4025fd1d2586139d24ce1dc/lib/waterline/model/lib/associationMethods/add.js#L131.  Meanwhile, `autoPK` can remain blissfully unawares of the new flag, except that it would set it at runtime on the generated attr def for the primary key.

sgress454:
This is no longer valid as `autoincrement` is now dealt with by the migration system in waterline-utils.  Thanks again @janhapke for the post!

#1314 is still an issue in Waterline 0.11.x, so that issue will stay open until Waterline 0.13.x is out of beta.

PR #1325, opened 2016-03-17
[Proposal] Pass additional argument to adapters

particlebanana:
Right now there doesn't exist a way to pass additional metadata for a query down to the adapter. This could be helpful when building custom adapters that need handle things like passing connections around or adding information for multi-tenancy. 

This PR adds that ability in a temporary format. For now you can tack on an additional argument to your queries after the callback or use the `.meta()` key in the deferred object builder to pass any arbitrary data down to an adapter function.

Yes this breaks the common node contract of the callback as the last argument and it's weird. I know this isn't a great longterm solution. If you don't use the extra argument everything will work as expected though. The big plus is it gives us this functionality now without breaking every single one of the community adapters. 

In the major version bump to `1.0.0` we will retool the function signatures to take an `(options, cb)` signature so that we can add non-breaking changes in the future.

Some examples of usage:

```javascript
// Callback Style
var meta = function() {
  return 'Hello, I\'m a meta function';
};

User.find({ name: 'tester' }, function(err, users) {
    // some logic
}, meta);
```

```javascript
// Deferred Style
var meta = function() {
  return 'Hello, I\'m a meta function';
};

User.find({ name: 'tester' })
.meta(meta)
.exec(function(err, users) {
 // some logic
});
```

To access the meta value in the adapter you would build a function like so:

```javascript
find: function(connectionName, table, options, cb, meta) {
  if(typeof meta === 'function') {
    console.log(meta());
  }
}
```

For some more background on this see https://github.com/balderdashy/waterline/pull/1307

### DIFF
--- a/lib/waterline/adapter/aggregateQueries.js
+++ b/lib/waterline/adapter/aggregateQueries.js
@@ -10,7 +10,7 @@ var hasOwnProperty = require('../utils/helpers').object.hasOwnProperty;
 module.exports = {
 
   // If an optimized createEach exists, use it, otherwise use an asynchronous loop with create()
-  createEach: function(valuesList, cb) {
+  createEach: function(valuesList, cb, metaContainer) {
     var self = this;
     var connName,
         adapter;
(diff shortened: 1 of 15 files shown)


Nishchit14:
:+1:    wow..
some hope for multi-tenancy.
 Any plan for **patch release** ?

particlebanana:
@Nishchit14 looking at getting this and the projections pr published as `0.12` today.

PR #1340, opened 2016-04-08
Added support for nested invalid attributes

Zaggen:
If you want to manually build an error based on `WLValidationError` that holds nested errors, because validation failed on nested data (association, nested documents, etc), there is no way to pass that data as invalidAttributes and get the expected result. There might not be support yet for deep validations but this opens the door for it. It's just a couple of simple helpers that iterate recursively if needed to properly normalize and format the errors.

So that this invalid attributes:

```js
{ location:
      { country:
         [ { rule: 'string',
             message: '`undefined` should be a string (instead of "null", which is a object)' },
           { rule: 'required',
             message: '"required" validation rule failed for input: null' },
           { rule: 'notEmpty',
             message: '"notEmpty" validation rule failed for input: null' } ],
        city:
         [ { rule: 'string',
             message: '`undefined` should be a string (instead of "null", which is a object)' },
           { rule: 'in',
             message: '"in" validation rule failed for input: null' },
           { rule: 'required',
             message: '"required" validation rule failed for input: null' },
           { rule: 'notEmpty',
             message: '"notEmpty" validation rule failed for input: null' } ],
        }
```
Can keep that format (otherwise it gets modified) and we can get a nice error too, like this
```
Error (E_VALIDATION)  ... some error details here
• location:
      • country:
        • `undefined` should be a string (instead of "null", which is a object)
        • "required" validation rule failed for input: null
        • "notEmpty" validation rule failed for input: null
      • city:
        • `undefined` should be a string (instead of "null", which is a object)
        • "in" validation rule failed for input: null
        • "required" validation rule failed for input: null
        • "notEmpty" validation rule failed for input: null
```

If the use case is not clear, or something needs to be improved or modified let me know.

pst: Not sure if there are tests for this, i didn't check since i'm doing this for the project i'm currently working on, so if there are some tests that needs to be modified somehow, let me know that too.

### DIFF
--- a/lib/waterline/error/WLValidationError.js
+++ b/lib/waterline/error/WLValidationError.js
@@ -49,29 +49,12 @@ function WLValidationError(properties) {
   this.model = properties.model;
 
   // Ensure messages exist for each invalidAttribute
-  this.invalidAttributes = _.mapValues(this.invalidAttributes, function(rules, attrName) {
-    return _.map(rules, function(rule) {
-      if (!rule.message) {
-        rule.message = util.format('A record with that `%s` already exists (`%s`).', attrName, rule.value);
-      }
-      return rule;
-    });
-  });
+  this.invalidAttributes = normalizeInvalidAttributes(this.invalidAttributes);
 
-  // Customize the `details`
-  this.details = util.format('Invalid attributes sent to %s:\n', this.model) +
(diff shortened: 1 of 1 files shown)


Zaggen:
ok, fixed!

PR #1343, opened 2016-04-15
[patch] Add fix to deal with buffers correctly in association cloning

philcaonz:
When populating, buffers are not cloned correctly with the default lodash cloneDeep function.


ftechz:
Fixed title!

PR #1351, opened 2016-05-10
check for NaN and inifinte values when coercing to integer

BrownBear2:
see https://github.com/balderdashy/sails/issues/3705


BrownBear2:
ok, fixed!

PR #1352, opened 2016-05-19
Run afterValidate with Model instance as `this`

scorpp:
Align context of `afterValidate` with those of `beforeValidate` and make `this` point to model instead of a global scope.

This is much beneficial since allows to implement common logic in single place (e.g. config/models.js in Sails).

Also it's just weird that `beforeValidate` and `afterValidate` which mirror each other are run in different contexts.


scorpp:
ok, fixed!

PR #1360, opened 2016-05-28
[patch] fixing update method so that it takes into account a string for the a…

jenjenut233:
…utoUpdatedAt attribute (was not updating the database record with the current timestamp)


jenjenut233:
fixed

joshuamarquez:
same problem here

particlebanana:
Thanks @jenjenut233!

particlebanana:
Published Waterline `0.11.3` to Waterline (for use with Sails) and will publish a new version of `0.12` in a bit.

PR #1392, opened 2016-09-09
[fix #1389] Problems with Through Associations with custom columnNames

### LINKED ISSUE #1389: Many-to-many association fails on save
**Waterline version**:0.11.0
**Node version**:4.4.2
**NPM version**:2.15.0
**Operating system**:Windows 10

<hr/>
**Sails version**: 0.12.3
**sails-mysql version**: 0.11.5

**Database**:
**MySQL**: 5.6

**Models:
User.js**

```
module.exports = {    
  attributes: {
    name: {
      type: 'string',
      size: 80,
      required: true
    },
    profiles:{
      collection: 'Profile',
      via: 'profile',
      through: 'userprofile'
    }
  }
};
```
**Profile.js**

```
module.exports = {   
  attributes: {
    type: {
      type: 'string',
      size: 50,
      required: true
    },
    users: {
      collection: 'User',
      via: 'user',
      through: 'userprofile'
    }
  }
};

```
**UserProfile.js**

```
module.exports = {
  tableName: 'user_profile',
  attributes: {
    profile: {
      model: 'Profile',
      columnName: 'profile_id'
    },   
    user:{
      model: 'User',
      columnName: 'user_id'
    }
  }
};
```
**Code:**

```
User.findOne({id: 1}).populate('profiles').exec(function (err, user) {
  user.profiles.add(2);
  user.save(function(err, saved){
  })
});
```
**Error:**

```
\node_modules\sails\node_modules\waterline\lib\waterline\model\lib\associationMethods\add.js:293
  var associationKey = collectionAttributes.attributes[attribute.on].via;
                                                                    ^

TypeError: Cannot read property 'via' of undefined

```
Thanks for any help.

pepeberba:
When dealing with through associations, waterlines' many-to-many fails when custom columnNames of the join table is defined. 

```
\node_modules\sails\node_modules\waterline\lib\waterline\model\lib\associationMethods\add.js:293
  var associationKey = collectionAttributes.attributes[attribute.on].via;
                                                                    ^

TypeError: Cannot read property 'via' of undefined
```

There seems to be a mix up with attribute.on and attribute.via, which is only different when the default columnNames are not used in a through association.

### DIFF
--- a/lib/waterline/model/lib/associationMethods/add.js
+++ b/lib/waterline/model/lib/associationMethods/add.js
@@ -290,17 +290,17 @@ Add.prototype.createManyToMany = function(collection, attribute, pk, key, cb) {
 
   // Grab the associated collection's primaryKey
   var collectionAttributes = this.collection.waterline.schema[attribute.collection.toLowerCase()];
-  var associationKey = collectionAttributes.attributes[attribute.on].via;
+  var associationKey = null;
 
   // If this is a throughTable, look into the meta data cache for what key to use
   if (collectionAttributes.throughTable) {
-    var cacheKey = collectionAttributes.throughTable[attribute.on + '.' + key];
+    var cacheKey = collectionAttributes.throughTable[attribute.via + '.' + key];
     if (!cacheKey) {
       return cb(new Error('Unable to find the proper cache key in the through table definition'));
     }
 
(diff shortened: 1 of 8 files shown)


pepeberba:
ok, fixed!

PR #1398, opened 2016-10-16
[fixes #1397]

### LINKED ISSUE #1397: toJSON error
Unhandled rejection TypeError: self.object[association].toJSON is not a function

all versions

calidion:
throw exceptions for all toJSON invocations.


calidion:
ok, fixed!

sgress454:
Not sure when the next v0.11.x release will be, and this bug is pretty obscure, but patched just in case.  Thanks for the report!

PR #1403, opened 2016-10-27
Fix aggregate options ignored in where

dinana:
If aggregate options are passed to where, they are saved to the criteria
object from normalize.criteria(criteria) however only criteria.where is
saved to _criteria.where, options saved to criteria object is ignored.


dinana:
ok, fixed!

PR #1407, opened 2016-11-10
Update async dependency and migrate code accordingly.

mikermcneil:


mikermcneil:
After the update:

```
∑ kit deps
  @sailshq/lodash@3.10.2       428.03 KB  0.214 seconds
  deep-diff@0.3.4              74.72 KB   0.037 seconds
  bluebird@3.2.1               581.24 KB  0.29 seconds
  flaverr@1.0.0                849.14 KB  0.424 seconds
  waterline-criteria@1.0.1     869.64 KB  0.434 seconds
  waterline-schema@0.2.0       913.68 KB  0.456 seconds
  switchback@2.0.1             877.02 KB  0.438 seconds
  async@2.0.1                  1.81 MB    0.909 seconds
  anchor@0.11.3  (~0.11.2)     3.74 MB    1.872 seconds
  prompt@1.0.0                 1.21 MB    0.609 seconds

Altogether, dependencies weigh in at 11.37 MB
(that will take an average of 5.68 seconds on coffee shop internet)

 (Note that `devDependencies` and `optionalDependencies` were NOT included above.)
```

Unfortunately, async@2 adds a bit of extra weight to the install.  But for most people, it's a difference of about 0.8 seconds, so not worth looking into further right now.  For more on that, see: https://github.com/caolan/async/issues/1283#issuecomment-259620172

PR #1412, opened 2016-12-07
Add a new example on how to initialize the waterline and expose the c…

elvisyang2016:
Add a new example on how to initialize the waterline and expose the collections.
Settle the problem caused by the waterline async initialization method.
Decouple the procedure of Waterline initialization with Express.


elvisyang2016:
ok, fixed!

PR #1417, opened 2016-12-19
[patch] Some changes so that samples match up

luislobo:
Some changes so that samples match up. Added comments prefixed by $$$, please review, and update accordingly or let me know.


luislobo:
ok, fixed!

mikermcneil:
@luislobo thanks!

PR #1419, opened 2016-12-21
[patch] Streamline validations

sgress454:
`sails-hook-rom` will now handle moving top-level properties like `isEmail` into a `validations` property on an attribute, so Waterline now just has to read that property instead of doing all the merging/checking itself.  Also, `required` isn't handled by anchor anymore so the code related to that has been removed.  And Waterline can just defer to anchor to complain about unknown validation rules.


sgress454:
Sorry @sailsbot !

PR #1424, opened 2016-12-29
[patch] Send model defaults to wl-schema

sgress454:
Companion to https://github.com/balderdashy/waterline-schema/pull/63


sgress454:
@sailsbot thanks bud

PR #1425, opened 2016-12-29
[patch] Always use column names for keys in s3 query joins

sgress454:


sgress454:
@particlebanana this is more complicated than i thought, let's look it over together and figure it out.  closing this mess for now.

PR #1426, opened 2017-01-04
[patch] Update help find

sgress454:
Ok, here is my attempt at flattening the operations runner into something a little easier to follow.  The gist is that it splits the "join" code into two paths -- one for when a junction table is used, one when it's not -- and either way, always runs 1 query (with user criteria, if present) on the child table per parent, rather than trying to do things like `skip` and `limit` in memory.  This is the least-optimized way to do non-native or cross-adapter joins, but it's the cleanest and we can add optimizations piecemeal while hopefully still maintaining some sense of order.

Easy optimization to start with would be going back to doing an `in` query for child records for a `to-one` association, since where/skip/limit/sort will never be relevant, so doing one query per parent is a waste.

Next would be doing an `in` query for child records on a `one-to-many` when where/skip/limit/sort are not used.

Both of those are fairly straightforward, trick is to do them while maintaining sanity.


sgress454:
Cool beans @sailsbot

PR #1431, opened 2017-01-10
[fixes #1429] One to many associations were being removed without checking that the actual related object

### LINKED ISSUE #1429: Association lost when using remove method
**Waterline version**:11.0
**Node version**:4.4.7
**NPM version**:2.15.9
**Operating system**:MACOS 10.12.2

<hr/>

I am currently working in a big project using sails.js and found the next issue. I was able to change my code, but I still think this a major waterline issue.

The problem shows when I have two models with a one to many association, If I overwrite the association it removes the old one. Which is right, but if I try to use the method remove for the array of the association already overwritten and then save the object, it removes the new one! I believe this happens because it only takes into account to id passed into the remove method and not the object calling it.

I made a small project to show the problem. With the models Player and Card. The function newRelationship in the player model shows the problem.
```
/**
 * Card.js
 *
 * @description :: TODO: You might write a short summary of how this model works and what it represents here.
 * @docs        :: http://sailsjs.org/documentation/concepts/models-and-orm/models
 */

module.exports = {

  attributes: {
    key: {
      type: "string",
      unique: true
    },

    player:{
      model: "Player"
    }

  }
};

```
```
/**
 * Player.js
 *
 * @description :: TODO: You might write a short summary of how this model works and what it represents here.
 * @docs        :: http://sailsjs.org/documentation/concepts/models-and-orm/models
 */

module.exports = {

  attributes: {
    name: {
      type: 'string',
      unique:true
    },

    cards: {
      collection: 'Card',
      via: 'player'
    },

    printWithCards: function(){
      var currentPlayer= this;
      Player.findOne(currentPlayer.id)
      .populate("cards")
      .exec(function(errDB,playerFound){
        console.log(playerFound);
      });
    }
  },

  newRelationship: function(cb){
    Player.create({name: "player1"}, function(errDB, player1){
      Card.create({key: "key1", player: player1}, function(errDB, card1){
        player1.printWithCards();//Player 1 with card1

        //Changing card to new player
        Player.create({name: "player2"}, function(errDB, player2){
          card1.player = player2;
          card1.save(function(errDB){
            //Changes made correctly, old relationship removed/overwritten
            player1.printWithCards();//Has no cards
            player2.printWithCards();//Has card 1

            //Removing old relationship wich does not exist
            player1.cards.remove(card1.id);
            //Saving after remove
            player1.save(function(errDB){
              player1.printWithCards(); //Player 1 has no card. Correct!
              player2.printWithCards(); //Player2 now has no cards. Wrong!!!
            });

          });
        });
      });
    });
  }
};
```

luislobo:
One to many associations were being removed without checking that the actual related object is the same as the one the user is trying to remove.
The fix adds a condition to the update, so that it makes sure that the value matches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 .dist
 coverage/
 npm-debug.log
+.idea
--- a/lib/waterline/model/lib/associationMethods/remove.js
+++ b/lib/waterline/model/lib/associationMethods/remove.js
@@ -43,7 +43,7 @@ var Remove = module.exports = function(collection, proto, records, cb) {
   //
   // In the future when transactions are available this will all be done on a single
   // connection and can be re-written.
-  this.removeCollectionAssociations(records, cb);
+  this.removeCollectionAssociations(records, proto, cb);
 };
 
 /**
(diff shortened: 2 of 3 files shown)


luislobo:
ok, fixed!

PR #1434, opened 2017-01-19
beforeFind

matt-cook:
Duplicate of https://github.com/balderdashy/waterline/pull/902 against current 0.10.x branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: node_js
 node_js:
-  - "0.12"
   - "0.10"
-  - "iojs"
+  - "0.12"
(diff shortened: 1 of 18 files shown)


matt-cook:
wrong repo.

mikermcneil:
@matt-cook Hey Matt, just posted in https://github.com/balderdashy/waterline/pull/1198#issuecomment-283238977 -- going to go through and cap off these other threads so that we can funnel discussion into one place.  (Also the link will cause them to get automatic back-references, so it'll be easier for folks from search engines to get up-to-date, relevant information this way)

PR #1438, opened 2017-01-25
Check adapter compatibility w/ Waterline 0.11.x

sgress454:


sgress454:
erm this doesn't look right

PR #1440, opened 2017-01-27
[patch] Log a warning when manually creating/updating timestamp values

sgress454:


mikermcneil:
Thanks!  Consolidated in https://github.com/balderdashy/waterline/commit/d9c922d402771fad50888935a0ea3886aad9fafc

...but that made me notice a weird side effect: when sails-hook-orm's automigrations run, we get:

image


So rather than introducing a new meta key to deal with this, I'm commenting this out and wrapping it in a "FUTURE: consider..." comment for now.  If we find that there is significant confusion being caused by the fact that providing an explicit value for an autoUpdatedAt attribute is now allowed, that justifies the added complexity of another meta key, then I'm open to revisiting this.

PR #1443, opened 2017-02-07
[Fixes #1442]

### LINKED ISSUE #1442: [Proposal] Nested Update's Should findOrCreate in some situations
**Waterline version**: 0.12.2
**Node version**: 6.6.0
**NPM version**: 4.1.1
**Operating system**: Ubuntu 16.04

<hr/>

Inside of `lib/waterline/query/dql/update.js` in the `createBelongsTo` function, you check whether the incoming data is an object but not whether it has any unique constraints before running the `.create` function.

If the incoming data contains an object at a related attribute (without specifying a value for the primary key) but the model specifies a unique constraint it should `findOrCreate`.

However, I wanted to submit a PR for this but the file no longer exists in the `master` branch.

I.E

```
models/people.js

{
  {
    name: {
      type: "string",
      unique: true
    },
    parent: {
      model: "people"
    }
  }
}
```

With the assumption that the following people exists in the DB already:

```
[
  {
    name: "dad"
  },
  {
    name: "son"
  }
]
```

If you run:

```
people.update({name: "son"}, {
  name: "son",
  parent: {
    name: "dad"
  }
}).then().blah()
```

It will error with `"0 attributes invalid"`. I've attached my file for your consideration, or if you can tell me a release branch I can clone and submit a PR to I'd be more than happy to help resolve the issue as I use this library in [Multicolour](/Multicolour/multicolour) and this affects the PUT handler (where any child item should be created or updated via a PUT request to be semantic)

[update.js.zip](https://github.com/balderdashy/waterline/files/755462/update.js.zip)

davemackintosh:
Fixes issue where object passed without primaryKey value present but has unique constraint. Don't always use the create method.

Fixes #1442

### DIFF
--- a/lib/waterline/query/dql/update.js
+++ b/lib/waterline/query/dql/update.js
@@ -120,18 +120,23 @@ function createBelongsTo(valuesObject, cb, metaContainer) {
     if (hop(attribute, 'model')) modelName = attribute.model;
     if (!modelName) return next();
 
+    var criteria = {};
     var model = self.waterline.collections[modelName];
     var pkValue = valuesObject.originalValues[item][model.primaryKey];
 
-    var criteria = {};
+    var isRelatedObjectWithoutPK = (hop(model.definition[item], "model") || hop(model.definition[item], "collection"));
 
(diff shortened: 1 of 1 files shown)


davemackintosh:
ok, fixed!

davemackintosh:
Ok, fixed!

goddamn.

PR #1445, opened 2017-02-07
[fixes #1444] Model.findOrCreate() should send back an object without {fetch: true} meta

### LINKED ISSUE #1444: Model.findOrCreate() should send back an object without {fetch: true} meta
**Waterline version**: 0.13.0-3
**Node version**: v7.5.0
**NPM version**: 4.1.2
**Operating system**: macOS 10.12.3 (16D32)

<hr/>

Hello,

I'm trying to use latest version of waterline with SailsJS 1.0.0-22, and found a problem with `Model.findOrCreate()`. It behaves the same as `Model.create()` rather than `Model.find()`, i.e it requires `{fetch: true}` meta to return a created/found object.

    Model.findOrCreate({ type: 'dog' }, { name: 'Pretend pet type', type: 'dog' }).then(console.log);
    // undefined

    Model.findOrCreate({ type: 'dog' }, { name: 'Pretend pet type', type: 'dog' }).meta({fetch: true}).then(console.log);
    // { id: 1, name: 'Pretend pet type', type: 'dog' }

mrded:
**Waterline version**: 0.13.0-3
**Node version**: v7.5.0
**NPM version**: 4.1.2
**Operating system**: macOS 10.12.3 (16D32)

<hr/>

Hello,

I'm trying to use latest version of waterline with SailsJS 1.0.0-22, and found a problem with `Model.findOrCreate()`. It behaves the same as `Model.create()` rather than `Model.find()`, i.e it requires `{fetch: true}` meta to return a created/found object.

    Model.findOrCreate({ type: 'dog' }, { name: 'Pretend pet type', type: 'dog' }).then(console.log);
    // undefined

    Model.findOrCreate({ type: 'dog' }, { name: 'Pretend pet type', type: 'dog' }).meta({fetch: true}).then(console.log);
    // { id: 1, name: 'Pretend pet type', type: 'dog' }

### DIFF
--- a/lib/waterline/methods/find-or-create.js
+++ b/lib/waterline/methods/find-or-create.js
@@ -251,6 +251,6 @@ module.exports = function findOrCreate( /* criteria?, newRecord?, done?, meta? *
       // > Note we set the `wasCreated` flag to `true` in this case.
       return done(undefined, createdRecord, true);
 
-    }, Object.assign({fetch: true}, query.meta));//</.create()>
+    }, _.extend({fetch: true}, query.meta));//</.create()>
   }, query.meta);//</.findOne()>
 };
--- a/lib/waterline/methods/find-or-create.js
+++ b/lib/waterline/methods/find-or-create.js
@@ -251,6 +251,6 @@ module.exports = function findOrCreate( /* criteria?, newRecord?, done?, meta? *
       // > Note we set the `wasCreated` flag to `true` in this case.
       return done(undefined, createdRecord, true);
 
-    }, query.meta);//</.create()>
+    }, Object.assign({fetch: true}, query.meta));//</.create()>
   }, query.meta);//</.findOne()>
 };
--- a/lib/waterline/methods/find-or-create.js
+++ b/lib/waterline/methods/find-or-create.js
@@ -251,6 +251,6 @@ module.exports = function findOrCreate( /* criteria?, newRecord?, done?, meta? *
       // > Note we set the `wasCreated` flag to `true` in this case.
       return done(undefined, createdRecord, true);
 
-    }, _.extend({fetch: true}, query.meta));//</.create()>
+    }, _.extend({fetch: true}, query.meta || {}));//</.create()>
   }, query.meta);//</.findOne()>
 };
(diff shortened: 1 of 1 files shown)


mrded:
ok, fixed!

mrded:
That's weird, because tests pass on my local env.

gil0mendes:
That is because the Object.assign isn't available in the older version of Node.js

mrded:
Oh right, I'll change it to something else. Thanks!

gil0mendes:
I think that is time to drop some version, but is just my opinion :)

mrded:
New node for new sails ;)

particlebanana:
@mrded good catch. We changed the behavior of `create` mid-stream and I forgot about the `findOrCreate` method.

mrded:
Looks like you have changed it already.

mikermcneil:
@mrded crap sorry about that, bad timing-- I just saw this.  For the purposes of tracking this for the future, we could still get this merged if you'd like

mrded:
It's ok. I would like to get it merged, let me solve conflicts first :)

mikermcneil:
@mrded and thanks for bringing it to our attention!!  Only way I even knew about this was because @particlebanana said it out loud- I just didn't realize there was already a PR

mrded:
@mikermcneil can you please check, what's a problem with tests?

mrded:
Will it be possible to merge it now? Otherwise it will never be merged :(

Thanks

mikermcneil:
@mrded totally-- I'm in there right now plugging in the optimized Deferred ([parley](http://npmjs.com/package/parley)) and I was actually just thinking about this and potential conflicts.  Thanks for the reminder!

mrded:
Awesome! Thanks

mrded:
Thank you for merging. Unfortunately you can see none of my changes in here https://github.com/balderdashy/waterline/commits/master/lib/waterline/methods/find-or-create.js

Looks like you merged my changes as empty commit. Anyway, problem is solved, and it's too late to complain.

Thanks!

mikermcneil:
@mrded hmm weird... I definitely merged this branch locally, fixed the conflict and then committed-- so maybe the commits aren't included in the commit history of the master branch because there was a merge conflict?  Git is a source of endless mysteries.  But regardless, it's in place now.

Thanks again for your help!

PR #1467, opened 2017-03-23
[patch] Fix issue where attribute is required and is a model attribute

luislobo:
Fix issue where attribute is required and is a model attribute. It was throwing an error about `Error: Consistency violation: `modelIdentity` must be a non-empty string.  Instead got: undefined` because collection was not existing, as it is a required model attribute.


luislobo:
ok, sailor, all fixed!

sgress454:
@luislobo Looks like @mikermcneil patched this awhile ago, but appreciate the contribution and Github says I can merge it anyway so...MERGED!

mikermcneil:
@luislobo 🍾  ⛵️  🏳️‍🌈

PR #1470, opened 2017-03-27
[patch] Removed repeated version comment for 0.11.4

luislobo:
Removed repeated version comment for 0.11.4


luislobo:
ok, fixed!

sgress454:
👍 Thanks @luislobo!  Looks like the changelog needs some updating in general... :-/

PR #1471, opened 2017-03-31
[patch] Remove dead code

ehmicky:
This commit removes an conditional branch that can never be triggered, because the same condition is checked few lines before, and throws an exception.


ehmicky:
ok, fixed!

sgress454:
👍 Thanks @ehmicky!

PR #1496, opened 2017-06-02
[proposal] Allow Date type in autoCreatedAt/autoUpdatedAt with the 'ref' type

Josebaseba:
Until the Sails1 version, the mongo adapter stored the createdAt/updatedAt values as Date types. So to keep everything working we need to allow the same Date type in the v1 of Sails.

With this PR using the 'ref' type will allow auto generate dates with the Date type:

```javascript
  attributes: {
    createdAt: { type: 'ref', autoCreatedAt: true },
    updatedAt: { type: 'ref', autoUpdatedAt: true },
    id: { type: 'string', columnName: '_id' }
  }
```
This PR goes with another PR in waterline-schema: https://github.com/balderdashy/waterline-schema/pull/90

Plus, we need to document this behavior that all the previous sails-mongo users will need (if they don't use the migrate: 'alter' option).

### DIFF
--- a/lib/waterline/utils/query/forge-stage-two-query.js
+++ b/lib/waterline/utils/query/forge-stage-two-query.js
@@ -1288,11 +1288,14 @@ module.exports = function forgeStageTwoQuery(query, orm) {
 
       // -• IWMIH, this is an attribute that has `autoUpdatedAt: true`,
       // and no value was explicitly provided for it.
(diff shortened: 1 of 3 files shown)


sgress454:
@mikermcneil This looks good to me -- allows backwards compatibility w/ earlier versions of ORM+sails-mongo that saved `createdAt` and `updatedAt` as native JS date objects, and would let postgresql and mysql users use `datetime` columns (w/ small updates to `sails-postgresql` and `sails-mysql`).

luislobo:
+100

luislobo:
Probably it would help if that snippet of code is added into: 
https://github.com/balderdashy/sails-generate/blob/f595e7787aa92805125950b8c68a2793a90468e5/lib/core-generators/new/templates/config/models.js#L53-L59

PR #1521, opened 2017-09-05
Soft deletes (work in progress)

mikermcneil:


luislobo:
@mikermcneil I like the `archive` idea. I see it could be great in some scenarios.
In our use case, we just need the soft-deleted models to be marked as "deleted" on the same instance, and filtered out, as we have history information that still needs to reference the "deleted" models.

mikermcneil:
@luislobo Glad to hear it.  And there's totally a place for an `isDeleted` flag too -- like sometimes you actually want to view the stuff in the UI, etc.  For our purposes here, `.archive()` is really more about just having something that you can use as a drop-in replacement for `.destroy()` when:

1. you know you have to keep the "deleted" data somewhere
2. but you also don't want it complicating things in your app if possible
3. And you don't need to be able to build any features involving the "deleted" data right away

luislobo:
@mikermcneil Yeah, I definitelly see the use and possibilities of archive. We might even use it instead of our current way of doing things. This aligned with some service that "autoarchives" records could be something that we might use.

PR #1528, opened 2017-09-27
[patch] Transform warnings state which model

wulfsolter:
Currently transform warnings to not state which model is the cause of the warning.

We've just done a 0.x -> 1.x Sails upgrade, and had a lot of nulls in places there shouldn't have been (one of the five transform warnings), but the warning does not state which model caused, just the property, which is insufficient when most models have properties like `createdAt` and `updatedAt`.


wulfsolter:
ok, fixed!

mikermcneil:
@wulfsolter 💯

PR #1529, opened 2017-10-02
This demonstrates how `mutateArgs` could be used to hypothetically make queries faster by avoiding unnecessary object cloning internally.  The next step is to benchmark this-- without benchmarks proving this is actually faster, it would be a mistake to merge this into core.

mikermcneil:


mikermcneil:
This demonstrates how `mutateArgs` could be used to hypothetically make queries faster by avoiding unnecessary object cloning internally.  The next step is to benchmark this-- without benchmarks proving this is actually faster, it would be a mistake to merge this into core.

mikermcneil:
@sgress454 I reckon you're talking about .archive() right?  (I think the PRs might have crossed streams)

sgress454:
@mikermcneil I was commenting in the source, I guess the specific commit must be part of this PR...

PR #1530, opened 2017-10-03
[proposal] encryption at rest

mikermcneil:
# TODOs

1. Finish up a few straggling things in this PR
2. Update docs (meta key is already covered, just need to write quite a bit more on this subject, including documenting the new model setting)
3. Have sails-generate set up new apps' `config/models.js` file with a default DEK for use during development, and update the boilerplate `config/env/production.js` file so that it explicitly wipes out that config to prevent people accidentally using the checked-in DEK in a production setting
4. Get this working in Node 4 (see https://github.com/simonratner/node-encrypted-attr/pull/3)
5. Figure out a workable solution for Node 0.10/0.12  (possibly just do a conditional require w/ a kind (but firm) error msg if you try to get fresh and fruity with at-rest encryption while using a "mature" version of node.js)

# New things

| What it is                                        | What kind of thing it is  | Data type              | Summary
|:---------------------------------|:------------------------|:------------------|:-----------|
| `dataEncryptionKeys`                  | Model setting                 | ((dictionary))         | A set of DEKs to use when encrypting and decrypting data.  The `default` DEK is used for all new encryption.  The other DEKs are there primarily to allow for decrypting older data that was encrypted with them before a key rotation.                  |
| `encrypt`                                       | Attribute prop                | ((boolean))            | Whether to auto-encrypt and (in most cases) auto-decrypt.   |
| `decrypt`                                      | Meta key                        | ((boolean))            | Whether to automatically decrypt records returned for this particular query.  Enable this flag to automatically decrypt data (otherwise, by default it'll come back still-encrypted.)                  |
| `encryptWith`                               | Meta key                        | ((string))            | The id of a custom key to use for encryption for this particular query.  (Otherwise by default, the `default` key is used for encryption.  For decryption, the appropriate key is always used based on the data being decrypted.)                   |

# Example

```js
// config/models.js
module.exports.models = {

  // In production, you'll want to set these using environment variables.
  // (e.g. `sails_models__dataEncryptionKeys__default='tVdQbq2JptoPp4oXGT94kKqF72iV0VKY/cnp7SjL7Ik='`)
  dataEncryptionKeys: {
    default: 'tVdQbq2JptoPp4oXGT94kKqF72iV0VKY/cnp7SjL7Ik=',
    '2016': 'DZ7MslaooGub3pS/0O734yeyPTAeZtd0Lrgeswwlt0s=',
    '2015': 'C5QAkA46HD9pK0m7293V2CzEVlJeSUXgwmxBAQVj+xU='
  },

};
```

```js
// api/models/User.js
module.exports = {
  
  // `dataEncryptionKeys` overrides could also be specified here on a per-model basis,
  // but realistically you won't ever need that in most cases.
 
  attributes: {

    fullName: {
      type: 'string',
      required: true,
      example: 'Maria Rogers'
    },

    ssn: {
      description: 'Social security number (encrypted at rest).',
      type: 'string',
      encrypt: true,
      example: '555-55-5555'
    },

  }
};
```

```js
// Usage:

await User.find();
// => 
// [ { id: 4, fullName: 'Wiggly Wriggly', ssn: 'YWVzLTI1Ni1nY20kJGRlZmF1bHQ=$F4Du3CAHtmUNk1pn$hMBezK3lwJ2BhOjZ$6as+eXnJDfBS54XVJgmPsg' } ]


await User.find().decrypt();
// => 
// [ { id: 4, fullName: 'Wiggly Wriggly', ssn: '555-55-5555' } ]
```

# Nice to have

Some kind of meta setting that lets you determine which key to use.  Normally, Waterline just uses the `default` DEK for encryption, and then whatever the appropriate DEK is for decryption.  So this would tell Waterline to use a different one of your DEKs when encrypting (e.g. `'2016'`).

This is handy during key rotation.  For example, consider the following code from a key rotation script:

```js
// assuming the same DEK is configured as both 'default' and '2016'...
//…

await User.stream()
.decrypt()
.eachRecord(async (user, next)=>{
  await User.update({ id: user.id }).set({ ssn: user.ssn }).meta({ encryptWith: '2016' });
  next();
});

// All social security numbers in the database have now been re-encrypted under
// the same DEK as before, but understanding it as key id '2016' instead of 'default'.
//
// That means that they can be decrypted normally now (regardless of whether this special
// meta key is present or not) as long as keys are configured for both 'default' and '2016'
// (since Waterline automatically figures out which key to use based on the key id embedded
// in the encrypted string.)  Remember, this `meta` setting only affects which DEK is used for
// _encryption_.   For decryption, Waterline always uses the appropriate DEK no matter what.
// 
// Finally, note that, at this point, any new encryption without providing the special meta flag
// will use 'default', as usual.  So if, after running this script, you configure a different key as
// 'default', then the key rotation will be complete!
```

##### Slightly more involved example

Here's the same example as above again, this time in a transaction:

```js
await User.getDatastore().transaction(async (db, proceed)=>{
  await User.stream().usingConnection(db)
  .decrypt()
  .eachRecord(async (user, next)=>{
    await User.update().usingConnection(db)
    .where({ id: user.id })
    .set({ ssn: user.ssn })
    .meta({ encryptWith: '2016' });
    return next();
  });
  return proceed();
});
```

### DIFF
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -196,6 +196,14 @@ function Waterline() {
                   });
                 }//•
 
+                if (attrDef.model || attrDef.collection) {
+                  throw flaverr({
+                    code: 'E_ATTR_NOT_COMPATIBLE_WITH_AT_REST_ENCRYPTION',
+                    attrName: attrName,
+                    whyNotCompatible: 'with associations.'
+                  });
+                }//•
+
                 if (attrDef.defaultsTo !== undefined) {
                   // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
                   // FUTURE: Consider adding support for this.  Will require some refactoring
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -158,10 +158,31 @@ function Waterline() {
       }
 
 
-      // Next, validate the `encrypt` attribute property, if it is in use.
+      // Next, validate ORM settings related to at-rest encryption, if it is in use.
       // =============================================================================================
+      _.each(wmds, function(wmd){
 
-      // TODO
+        var modelDef = wmd.prototype;
+
+        // Verify that `encrypt` attr prop is valid, if in use.
+        _.each(modelDef.attributes, function(attrDef, attrName){
+          if (attrDef.encrypt !== undefined) {
+            if (!_.isBoolean(attrDef.encrypt)){
+              throw new Error(
+                'Invalid usage of `encrypt` in the definition of `'+modelDef.identity+'` model\'s '+
+                '`'+attrName+'` attribute.  If set, `encrypt` must be either `true` or `false`.'
+              );
+            }//•
+            if (attrDef.encrypt === true && attrDef.type === 'ref') {
+              throw new Error(
+                'Invalid usage of `encrypt` in the definition of `'+modelDef.identity+'` model\'s '+
+                '`'+attrName+'` attribute.   At-rest encryption (`encrypt: true`) cannot be used '+
+                'with `type: \'ref\'` attributes.'
+              );
+            }//•
+          }//ﬁ
+        });//∞
+      });//∞
 
 
       // Next, set up support for the default archive, and validate related settings:
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -167,24 +167,75 @@ function Waterline() {
         var modelDef = wmd.prototype;
 
         // Verify that `encrypt` attr prop is valid, if in use.
-        _.each(modelDef.attributes, function(attrDef, attrName){
-          if (attrDef.encrypt !== undefined) {
-            if (!_.isBoolean(attrDef.encrypt)){
-              throw new Error(
-                'Invalid usage of `encrypt` in the definition of `'+modelDef.identity+'` model\'s '+
-                '`'+attrName+'` attribute.  If set, `encrypt` must be either `true` or `false`.'
-              );
-            }//•
-            if (attrDef.encrypt === true && attrDef.type === 'ref') {
-              throw new Error(
-                'Invalid usage of `encrypt` in the definition of `'+modelDef.identity+'` model\'s '+
-                '`'+attrName+'` attribute.   At-rest encryption (`encrypt: true`) cannot be used '+
-                'with `type: \'ref\'` attributes.'
-              );
-            }//•
-          }//ﬁ
-        });//∞
+        try {
+          _.each(modelDef.attributes, function(attrDef, attrName){
+            if (attrDef.encrypt !== undefined) {
+              if (!_.isBoolean(attrDef.encrypt)){
+                throw flaverr({
+                  code: 'E_INVALID_ENCRYPT',
+                  attrName: attrName,
+                  message: 'If set, `encrypt` must be either `true` or `false`.'
+                });
+              }//•
 
+              if (attrDef.encrypt === true){
+
+                if (attrDef.type === 'ref') {
+                  throw flaverr({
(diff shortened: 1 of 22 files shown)


mikermcneil:
> I, for instance, would prefer to decrypt only when requested with a meta key, so that I specifically have to tell it to decrypt the data

@luislobo That sounds reasonable to me.  How about a model setting instead?  Seems like a preference that won't change on a query by query basis

mikermcneil:
@luislobo actually man after some further consideration, I totally agree with your behavior.  That makes way more sense as a default.  I edited the notes in the PR above to reflect that and I'm working on the implementation now.

mikermcneil:
@luislobo ok so I just realized a problem with this.  If we don't automatically decrypt data by default, since we still _encrypt_ it by default, then you have to keep track of whether you've decrypted it in your code (and inevitably you forget, meaning you end up with double-encrypted data)

mikermcneil:
Well.. after a cursory test this is actually fine.  It broke "alter" automigrations, but I fixed and published that already.  This way there's the occasional bug you're going to run into when building apps-- but better that than leaking people's social security numbers

mikermcneil:
Alright, as far as I see it, this is ready to merge.  It breaks Node 4 compat., but there's a plan in place to solve that in node-encrypted-attr.  If there's any holdups there, we can always just make a fork that's compatible with node 4  (and might as well try for node 0.10, node 0.12 while at it, if possible)

luislobo:
FWIW, I checked encrypting an already encrypted object, and they return the same thing
```javascript
const crypto = require('crypto')
const EncryptedAttributes = require('encrypted-attr')


let options = {
      keys: {
        k1: 'EMr5G6mKU03yHjc8gI1eutCyRs4zl+Q9xpSUdKpElZY=',
        k2: 'Yuk4t505JuB7xeQfHgv6RZJZcQg+P3eTYkB7IRutOS8='
      },
      keyId: 'k1',
      verifyId: true
    }

let enc = EncryptedAttributes(['secret'], options)
    let obj = {id: 1, secret: 'value'}
    // expect: YWVzLTI1Ni1nY20kMSRrMQ==$sK91YfUvv+O8Jx/m$OOQniq8=$WLbWYz7uCQBTNO3Fc+5UvA

    let eo = enc.encryptAll(obj);
    let eo2 = enc.encryptAll(eo);

    console.log(eo)
    console.log(eo2)
```

luislobo:
RE: node 4 compatibility, I could not make encrypted-attr work on Node 4 at all, even transpiling the code, so, that might be a huge deal.

mikermcneil:
@luislobo yeah I wish the stragglers would drop 0.x, but that's ok-- I know what it's like to be a luddite :p  

Thanks for looking into it-- as it turns out, node 0.x didn't support the right stuff in the native crypto module.  I did a PR that adds a little check up top.  Basically, since it's a new feature, as long as at-rest encryption works in node 8, I'm happy... as long as it doesn't break anything else  in 0.10+, and as long as, if you try to use at-rest encryption in an unsupported Node version, it yells at you in an easy-to-understand-and-resolve way

mikermcneil:
@luislobo well looks like it's supposed to be in node 4's `crypto`: https://github.com/meteor/meteor/issues/4461#issuecomment-236850175

So once all the syntax sugar is sucked out of it, this should be working in node 4 .... (but the exact usage of `crypto` could have changed between 4 and 6 I guess-- so maybe that?)

luislobo:
Yeah, it's not about crypto not being supported but the way it is implemented in that module. May be using another module with similar characteristics might work better

grantbi:
@mikermcneil is there a way, or could we have a way to migrate an existing db with this encryption-at-rest method? When I lift sails (without dropping) it fails to decrypt and will not continue.

I would appreciate any suggestions!

PR #1531, opened 2017-10-03
[patch] Add capitalized `globalId` to default archive model.

sgress454:
For consistency with expectation that model globals are capitalized (without this, the global is `archive`).

### DIFF
--- a/lib/waterline.js
+++ b/lib/waterline.js
@@ -268,6 +268,7 @@ function Waterline() {
 
           var newWmd = Waterline.Model.extend({
             identity: DEFAULT_ARCHIVE_MODEL_IDENTITY,
+            globalId: _.capitalize(DEFAULT_ARCHIVE_MODEL_IDENTITY),
             primaryKey: 'id',
             datastore: arbitraryArchiver.datastore,
             attributes: {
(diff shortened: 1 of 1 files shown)


sgress454:
@sailsbot ❤️

mikermcneil:
@sgress454 Hmm good catch-- and actually this is true for implicit join models too...  Currently, Waterline isn't aware of `globalId` anywhere else (the globalization and globalId-related stuff is in include-all+sails core) so I'd rather avoid including it here for consistency and to avoid future confusion.

But it begs the question of how we can allow a model _not_ to become globalized.  Maybe a new model setting?  But it'd be better to make it implementation agnostic  (also to continue to keep the whole globals thing out of waterline core).

How about `private: true`?

mikermcneil:
Or better yet, `_private`.  If ever someone finds the need for customizing this behavior in userland, then we can discuss making it officially supported- but until then, it's for internal use.

https://github.com/balderdashy/sails-hook-orm/commit/650eec73d78d6aca58edadb7216ad2f2229c6578


As for whether to expose an Archive global by default, on second thought, I think you're probably right.  If you have globals enabled in Sails, you'd expect to be able to easily use it so it makes sense.  Once I've made the other change in wl-schema I'll come back, test this out, then merge

PR #1532, opened 2017-10-13
[patch] Use Node version check to determine whether to process encryption options for each model

sgress454:


sgress454:
@sailsbot I dream of thee

sgress454:
@mikermcneil the diff looks terrible but it's really just wrapping that whole block of code in 

```
if (isNativeCryptoFullyCapable) {
```

PR #1545, opened 2018-01-09
[fixes #1509] Fail normalizeWhereClause() for undefined attributes configurable

### LINKED ISSUE #1509: Find, Destroy or Update methods ignore undefined type attributes
**Waterline version**: 0.13.0-rc11
**Node version**: 4.5
**NPM version**: 2.15.9
**Operating system**: MacOs

<hr/>

Hi guys, I found an issue that I'm not sure if it's a bug or something that you have considered. If you do something like this:

```javascript
User.update({id: undefined}, {name: 'joe'}).exec()
```

This is gonna update ALL the users in the database, because the mongo query is gonna be something like this {}. The same thing happens with find/destroy methods.

So imagine that in your code you don't realize data X value is undefined and you lauch a destroy against it, you'll end up dropping all the collection with no clue about what just happened.

Thanks for your time.

davidreghay:
If failOnUndefinedAttributesInWhereClause is set on a given Model,
then providing keys in a `where` clause with `undefined` as values
fails with E_WHERE_CLAUSE_UNUSABLE. The query.find.js test suite
has been updated to reflect this change.

### DIFF
--- a/lib/waterline/utils/query/private/normalize-where-clause.js
+++ b/lib/waterline/utils/query/private/normalize-where-clause.js
@@ -304,6 +304,12 @@ module.exports = function normalizeWhereClause(whereClause, modelIdentity, orm, 
       // Strip out any keys with undefined values.
       _.each(_.keys(branch), function (key){
         if (_.isUndefined(branch[key])) {
+          if(WLModel.failOnUndefinedAttributesInWhereClause) {
+            throw flaverr('E_WHERE_CLAUSE_UNUSABLE', new Error(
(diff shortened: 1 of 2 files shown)


davidreghay:
ok, fixed!

davidreghay:
@mikermcneil @sgress454 , I know you've got your hands full with other stuff but this seems like low-hanging fruit that would be really appreciated by sails users bumping up against https://github.com/balderdashy/waterline/issues/1509

davidreghay:
Looks like this isn't needed

PR #1579, opened 2018-10-17
Allow 0 to be a valid UNIX timestamp.

falconscript:
Issuing console warnings for timestamp of 0 isn't necessary and are sometimes used to represent "unset" programmatically since datetimes must be type 'ref' and 'ref' cannot be null. Removing the if check will allow them to pass through in auto-migrations without spurring messages.


falconscript:
[proposal] Added

PR #1584, opened 2018-11-08
[patch] Fix incorrect link in README.md

shockey:
Unclear where the original link came from - maybe a rogue copy-paste? 🤔 

PS: thanks for building Waterline, it's a wonderful library to work with!


shockey:
ok, fixed!

PR #1587, opened 2018-12-07
PROPOSAL: Update transformer-builder

lelkadi:
In the unserialize(), only transform pRecord's columns that match model's attributes. 
Or in other words, skip any fields available in the pRecord that don't match any attribute of the Model.


lelkadi:
ok, fixed!

lelkadi:
I think this proposed change is essential for an ORM, as it allows more control on the data on the Model level.

PR #1598, opened 2019-03-22
Don't ignore undefined keys in where clauses

alxndrsn:
Fixes https://github.com/balderdashy/sails/issues/4639

PR #1599, opened 2019-03-22
fixes #4639 Throw Error on undefined value in WHERE clause

alxndrsn:
Fixes https://github.com/balderdashy/sails/issues/4639

Alternative to #1598


alxndrsn:
@sailsbot

alxndrsn:
@sailsbot

PR #1600, opened 2019-04-26
Add more info to error messages when using min/max as column names

alxndrsn:


rachaelshaw:
@alxndrsn Thanks for helping improve these error messages! I went ahead and made a couple changes before merging:
• Moved the additional text below the example code in the error messages, just so people don't get confused about what the example is demonstrating (i.e. so they don't assume it's an example of things not working as expected)
• Added the same message for `groupBy`, `sum`, and `average`, since those have the same problem

PR #1606, opened 2019-05-17
Fixes sails/#4737

alxndrsn:
Remove linebreaks from long error messages

### DIFF
--- a/lib/waterline/utils/query/process-all-records.js
+++ b/lib/waterline/utils/query/process-all-records.js
@@ -17,27 +17,23 @@ var eachRecordDeep = require('waterline-utils').eachRecordDeep;
 var WARNING_SUFFIXES = {
 
   MIGHT_BE_YOUR_FAULT:
-  '\n'+
-  '> You are seeing this warning because there are records in your database that don\'t\n'+
-  '> match up with your models.  This is often the result of a model definition being\n'+
-  '> changed without also migrating leftover data.  But it could also be because records\n'+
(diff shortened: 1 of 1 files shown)


lib/waterline/utils/query/process-all-records.js
```diff
@@ -228,15 +224,15 @@ module.exports = function processAllRecords(records, meta, modelIdentity, orm) {
...
-              '```\n'+
-              util.inspect(record, {depth:5})+'\n'+
-              '```\n'+
+            console.warn(
+              'Warning: Records sent back from a database adapter should always have a valid property '+
+              'that corresponds with the primary key attribute (`'+WLModel.primaryKey+'`).  But in this result set, '+
+              'after transforming columnNames back to attribute names for model `' + modelIdentity + '`, '+
+              'there is a record with a missing or invalid `'+WLModel.primaryKey+'`. '+
+              'Record: '+
+              '``` '+
```
rachaelshaw: @alxndrsn `util.inspect` is going to add new lines here anyway, so I think in spots like this it may just make more sense to leave the lines before and after the backticks. Any thoughts on this?

alxndrsn: :man_shrugging: in which case I should probably just override `console.log`, `console.error` etc. :confused:

wulfsolter:
Changing content of error messages will break tests that have expectations of a certain response/error.

This PR will add a non-zero amount of work, for what I can only interpret as a cosmetic change. Just my 2c though.

PR #1615, opened 2020-04-09
[patch] Allow populates to use select/omit for singular associations

danielsharvey:
Currently the populates clause is allowed for plural ("collection")
associations but disallowed for singular ("model") associations.

This patch allows only a select or an omit for singular associations.


danielsharvey:
I have review the code, run the tests and tested within my own use cases.

Are there other tests that should be run?

PR #1616, opened 2020-05-29
Fix .archive() and .archiveOne() when using custom column names

mikermcneil:
Fix bug that was occuring in some cases when using .archive() and .archiveOne() with custom column names; including w/ sails-mongo or sails-disk in the new mongo mode


mikermcneil:
released in `waterline@0.14.0-5`

PR #1619, opened 2021-04-20
Run hooks with Model instance as this

sivo1981:
Implemented option to call hooks 'beforeUpdate', 'afterUpdate', 'beforeCreate', 'afterCreate', 'beforeDestroy', 'afterDestroy', with Model instance and make this point to model instead of a global scope.

This allows to implement common logic in single place (e.g. config/models.js in Sails).


sivo1981:
Implements [https://github.com/balderdashy/sails/issues/7113](url)

PR #1620, opened 2021-08-16
Update url in readme

alexjv89:
The previous url is pointing to a different section in the documentation.


eashaw:
Thanks for catching that @alexjv89

PR #1621, opened 2021-09-28
check datastore connectivity after registering it

bhargavkn:
Changes in this PR attempt to resolve balderdashy/sails#7095 by making waterline `validateConnection` (on adapters that implement the method) for a datastore after it has been registered. This will make waterline return an error if a datastore was registered but waterline's attempt to validate the connection failed.


bhargavkn:
@eashaw may I please have your review on this PR? 🙇

lib/waterline.js
```diff
@@ -711,7 +711,19 @@ function Waterline() {
...
 
         // Call the `registerDatastore` adapter method.
-        datastore.adapter.registerDatastore(datastore.config, usedSchemas, next);
+        datastore.adapter.registerDatastore(datastore.config, usedSchemas, function registerDatastoreCb(err) {
+          if (err) {
+            return next(err);
+          }
+
+          // Check if the datastore adapter has a `validateConnection` method
+          if (_.has(datastore.adapter, 'validateConnection')) {
```
eashaw, line 720: Could you instead use `getConnection` and `releaseConnection` adapter methods (you'll still need to check if they both exist)
Heres an example of how you can access the `manager`
https://github.com/balderdashy/sails-hook-orm/blob/9c2a3c768eba152e47af645d02462ed831f87dbb/lib/build-registered-datastore-instance.js#L116

eashaw (changes requested):

lib/waterline.js
bhargavkn: @eashaw The idea here was to abstract the details of things such as `getConnection` and `releaseConnection` (which are more common in the SQL realm and might not be applicable in other cases) away from waterline and keep it contained within the adapters (as done in https://github.com/balderdashy/sails-postgresql/pull/290). This way each adapter can be more flexible in terms of how it wants to validate a connection and waterline needn't be concerned with the implementation details of how connectivity to a datastore is to be validated. This is my opinion though, Let me know if you still want me to change this.

PR #1627, opened 2025-04-11
feat: improve forgeStageTwoQuery to allow select/omit for singular model populate

### LINKED ISSUE #4667: Ambiguous usage with singular model association
**Sails version**: 1.1.0
**Node version**: 8.10
**NPM version**: 5.6.0
**DB adapter name**: sails-mysql
**DB adapter version**: 1.0.1
**Operating system**: Debian GNU/Linux 8 (jessie)

<hr/>
I'm running my project on docker image with *nodev8.10*. I have a model with a single association with another model, example

    /*Pet*/
    module.exports = {
        attributes: {
          name: { type: 'string' },
          petType: {
            model: 'pettype'
          }
        }
    }

    /*PetType*/
    module.exports = {
        attributes: {
          name: { type: 'string' },
        }
    }

Then I try to populate my Pet model with specific PetType fields, like this:

    await Pet.find().populate('petType', { select: ['name'] });

And I get the following error:

> Could not populate `petType` because of ambiguous usage.  This is a singular ("model") association, which means it never refers to more than _one_ associated record.  So passing in subcriteria (i.e. as the second argument to `.populate()`) is not supported for this association, since it generally wouldn\'t make any sense.  But that\'s the trouble-- it looks like some sort of a subcriteria (or something) _was_ provided!\n(Note that subcriterias consisting ONLY of `omit` or `select` are a special case that _does_ make sense.  This usage will be supported in a future version of Waterline.)\n\nHere\'s what was passed in:\n{ select: [ \'name\' ] }

DominusKelvin:
Resolves https://github.com/balderdashy/sails/issues/4667

I also account for the normal behaviour of Waterline projections not allowing both `select` and `omit`

CleanShot 2025-04-11 at 14 58 44@2x

Now this sort of subcriteria will be treated as valid subcriteria for populating associations

```
 const tickets = await Ticket.find({
      where: { event: event.id, status: { '!=': 'cancelled' } },
      sort: 'ticketHolderEmail DESC',
      select: ['ticketHolderEmail', 'status', 'publicId']
    })
      .populate('ticketType', { select: ['name'] })
      .populate('order')
```


luislobo:
Good improvement! It would be good to have unit tests

DominusKelvin:
Yes @luislobo

DominusKelvin:
Hey @luislobo. I just added three test cases for this.

test/unit/query/associations/populateArray.js
```diff
@@ -123,5 +123,49 @@ describe('Collection Query ::', function() {
...
+      Car.find()
+      .populate('driver', {
+        select: ['name']
+      })
+      .exec(function(err, car) {
+        if (err) {
+          return done(err);
+        }
+
+        assert(car[0].driver);
```
luislobo, line 137: Another assert should check that name is there and no other fields are present

test/unit/query/associations/populateArray.js
```diff
@@ -123,5 +123,49 @@ describe('Collection Query ::', function() {
...
+      Car.find()
+      .populate('driver', {
+        omit: ['id']
+      })
+      .exec(function(err, car) {
+        if (err) {
+          return done(err);
+        }
+
+        assert(car[0].driver);
```
luislobo, line 153: Another assert should check that there are other fields present, but not id

test/unit/query/associations/populateArray.js
```diff
@@ -123,5 +123,49 @@ describe('Collection Query ::', function() {
...
+        select: ['name'],
+        omit: ['id']
+      })
+      .exec(function(err, car) {
+        assert(err);
+        assert.equal(err.code, 'E_INVALID_POPULATES');
+        assert(err.message.indexOf('providing both `select` AND `omit`') > -1);
+        return done();
+      });
+    });
```
luislobo, line 185: what about this case?

```javascript
Car.find()
      .populate('driver',{})
```

luislobo:
@DominusKelvin great! I added some comments

test/unit/query/associations/populateArray.js
DominusKelvin: Hey @luislobo I don't think Waterline projections a.k.a select/omit allows you to omit the `id` field.

test/unit/query/associations/populateArray.js
DominusKelvin: Done

test/unit/query/associations/populateArray.js
luislobo: Then, you should change the test to not use id for omit, and test that the field you chose is omited.

test/unit/query/associations/populateArray.js
luislobo: This test is not enough, you should test that the driver only has name.
When I tested locally, it also had `car:1`

Image

Add this to your test and you will see that it is not only selecting name but bringing other attributes:

```
        console.log(car[0].driver);
        assert(car[0].driver);
        assert(car[0].driver.name);
        assert(car[0].driver.id);
        assert(Object.keys(car[0].driver).length === 2);
```

luislobo:
I've added more comments, there is actually an issue on "select"

test/unit/query/associations/populateArray.js
luislobo: Similar issue here, if you add this, the test doesn't pass:

```javascript
        console.log(car[0].driver);
        assert(car[0].driver);
        assert(car[0].driver.name);
        assert(car[0].driver.car === undefined);

```

DominusKelvin:
> I've added more comments, there is actually an issue on "select"

Hey @luislobo what's the issue you saw?

luislobo:
your changes do not do what the issue is asking: omit does not omit and select does not select, because the current waterline implementation does not process select and omit on populates, it only does it on the main model's select and omit. Your change just lets those fields to be validated in the configruation. If you add the tests that I mentioned on the code, you will see what I mean.

luislobo:
BTW, I do have an implementation for this version of Waterline that implements this, I could create a separate PR that does it. But since I'm not using this version of Waterline yet, I didn't create a PR.

DominusKelvin:
> BTW, I do have an implementation for this version of Waterline that implements this, I could create a separate PR that does it. But since I'm not using this version of Waterline yet, I didn't create a PR.

Oh really I did test it out on my current Sails and Waterline versions

I'll test it out again to confirm

luislobo:
Here's the PR that takes care of the select/omit on propulated entities https://github.com/balderdashy/waterline/pull/1628

PR #1628, opened 2025-09-24
feat: support select/omit projections on populates

luislobo:
## Summary

- teach  to snapshot the original populate instructions and apply the requested  /  projection after unserializing populated child records; this keeps adapter quirks from leaking extra attributes
- allow singular () populates to specify  / ; we validate the shapes, require array syntax, auto-include the child PK, reject unknown / collection attributes, and block custom projections when the associated model is 
- build out regression coverage for the happy path and every new  error branch (bad types, unknown fields, collection / PK references, schema:false models), derived from sailscastshq/waterline@414c927590893a5eefa632f92cbed1f0b3d75d42

## Why

Waterline already supports  /  on top-level criteria, but populated singular associations silently ignored those directives. Adapters that don’t implement column-level projections returned fully hydrated child rows, so API users had no reliable way to trim sensitive attributes out of populated associations.
The validation code also rejected otherwise reasonable populate subcriteria, making it impossible to express  /  even if we applied them ourselves.

This change brings singular populates up to parity with top-level projections, keeps adapters honest, and documents the exact set of error cases developers should expect when they misuse the feature.

## How

1. Capture the original  dictionary in , then run a tiny projector over every populated child record once we’ve transformed column names back into attribute names.
2. Extend the singular-attribute normalization branch in  so that:
   - only , , or  are accepted
   - the arrays contain strings that resolve to real attributes on the associated model
   - the associated model’s primary key remains selected
   - schema:false associated models continue to forbid custom projections
3. Port the sailscastshq test additions and expand them to exercise every failure mode (non-arrays, non-strings, unknown attributes, collection attributes, schema:false usage, PK omission).

## Testing

- npm test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Edge
 
 ##### General
+* [ENHANCEMENT] Allow `select`/`omit` projections on populated singular associations (ported from sailscastshq/waterline@414c927).
 * [BUGFIX] Fix .archive() and .archiveOne() when using custom column names  (#1616)
 * [BREAKING] Waterline attribute names must now be [ECMAScript 5.1-compatible variable names](https://github.com/mikermcneil/machinepack-javascript/blob/3786c05388cf49220a6d3b6dbbc1d80312d247ec/machines/validate-varname.js#L41).
   + Custom column names can still be configured to anything, as long as it is supported by the underlying database.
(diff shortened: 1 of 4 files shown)


luislobo:
Resolves https://github.com/balderdashy/sails/issues/4667

